PR #70, opened 2014-10-23
Consider object and array patterns for scope

fkling:
Currently, using an object pattern like

    var {foo, bar} = baz;

does not add `foo` and `bar` to the scope.

This change adds support for these patterns and extends the current type definitions to make them compatible with esprima. I hope this OK.

I believe there is a deviation from the Mozilla Parser API. Even though `PropertyPattern` is not officially defined, its signature in [`ObjectPattern`](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Parser_API#Patterns) looks like `{ key: ..., value: ... }` whereas ast-types uses `{key: ..., pattern: ... }`.

### DIFF
--- a/def/es6.js
+++ b/def/es6.js
@@ -103,6 +103,14 @@ def("SpreadElementPattern")
     .build("argument")
     .field("argument", def("Pattern"));
 
+def("ArrayPattern")
+    .field("elements", [or(
+      def("Pattern"),
+      null,
+      // used by esprima
+      def("SpreadElement")
+    )]);
+
 var ClassBodyElement = or(
     def("MethodDefinition"),
     def("VariableDeclarator"),
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -225,10 +225,36 @@ function addPattern(patternPath, bindings) {
         } else {
             bindings[pattern.name] = [patternPath];
         }
+    } else if (namedTypes.ObjectPattern.check(pattern)) {
 
-    } else if (namedTypes.SpreadElement &&
-               namedTypes.SpreadElement.check(pattern)) {
-        addPattern(patternPath.get("argument"), bindings);
+        patternPath.get('properties').each(function(propertyPath) {
+            var property = propertyPath.value;
+            if (namedTypes.Pattern.check(property)) {
+                addPattern(propertyPath, bindings);
+            } else  if (namedTypes.Property.check(property)) {
+                addPattern(propertyPath.get('value'), bindings);
+            } else if (namedTypes.SpreadProperty &&
+                       namedTypes.SpreadProperty.check(property)) {
+                addPattern(propertyPath.get('argument'), bindings);
+            }
+        });
+    } else if (namedTypes.ArrayPattern.check(pattern)) {
+        patternPath.get('elements').each(function(elementPath) {
+            var element = elementPath.value;
+            if (namedTypes.Pattern.check(element)) {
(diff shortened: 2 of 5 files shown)


fkling:
Sorry for letting this sit for so long. Regarding "aliasing" `.pattern` and `.value`, I don't think there is much value to get from this, unless we make `Property` a `Pattern` as well.

fkling:
Ping?

def/es6.js
```diff
@@ -97,6 +97,16 @@ def("SpreadElementPattern")
...
     .field("argument", def("Pattern"));
 
+def("ArrayPattern")
+    .bases("Pattern")
+    .build("elements")
+    .field("elements", [or(
+      def("Pattern"),
+      null,
+      // used by esprima
+      def("SpreadElement")
```
benjamn: Excellent point.

lib/scope.js
```diff
@@ -225,10 +225,36 @@ function addPattern(patternPath, bindings) {
...
+            if (namedTypes.Pattern.check(property)) {
+                addPattern(propertyPath, bindings);
+            } else  if (namedTypes.Property.check(property)) {
+                addPattern(propertyPath.get('value'), bindings);
+            } else if (namedTypes.SpreadProperty &&
+                       namedTypes.SpreadProperty.check(property)) {
+                addPattern(propertyPath.get('argument'), bindings);
+            }
+        });
+    } else if (namedTypes.ArrayPattern.check(pattern)) {
```
benjamn: Nit: `ArrayPattern` is only defined in `def/es6.js`, so can you guard this condition like so?
```js
} else if (namedTypes.ArrayPattern &&
           namedTypes.ArrayPattern.check(pattern) {
```

def/es6.js
```diff
@@ -103,6 +103,14 @@ def("SpreadElementPattern")
     .build("argument")
     .field("argument", def("Pattern"));
 
+def("ArrayPattern")
+    .field("elements", [or(
+      def("Pattern"),
```
benjamn: Nit: 4 spaces for indentation plz

lib/scope.js
```diff
@@ -225,10 +225,36 @@ function addPattern(patternPath, bindings) {
...
+        patternPath.get('elements').each(function(elementPath) {
+            var element = elementPath.value;
+            if (namedTypes.Pattern.check(element)) {
+              addPattern(elementPath, bindings);
```
benjamn: Indentation.

test/run.js
```diff
@@ -1016,6 +1016,133 @@ describe("catch block scope", function() {
...
+            assert.strictEqual(scope.declares("b"), false);
+            assert.strictEqual(scope.declares("c"), true);
+            assert.strictEqual(scope.declares("d"), true);
+            assert.strictEqual(scope.declares("e"), false);
+            assert.strictEqual(scope.declares("foo"), true);
+            assert.strictEqual(scope.declares("bar"), true);
+            assert.strictEqual(scope.declares("baz"), true);
+        });
+    });
+});
```
benjamn: Thanks for all the tests!

benjamn:
I can easily fix the couple of nits that I had, so I'm going to merge this as-is :)

fkling:
Thank you so much! I switched our React documentation parser (which will soon be open sourced) to use Recast and this will allow me to do some more advanced analyses.

PR #86, opened 2015-01-22
Add definition for ArrayTypeAnnotation

gabelevi:
I neglected to add a definition for ArrayTypeAnnotation, which is the type of the `number[]` syntax, like

  function foo(x: number[]) {}


benjamn:
Published NPM version 0.6.10 with this change.

PR #87, opened 2015-02-04
Add definition for TypeCastExpression

gabelevi:
@bhosmer has added this pretty neat feature to Flow called type casts. They look like 
`(expression: type)`.

### DIFF
--- a/def/fb-harmony.js
+++ b/def/fb-harmony.js
@@ -158,7 +158,7 @@ def("ArrayTypeAnnotation")
 
 def("ObjectTypeAnnotation")
   .bases("Type")
-  .build("properties")Al
+  .build("properties")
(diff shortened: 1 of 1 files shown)


def/fb-harmony.js
```diff
@@ -158,7 +158,7 @@ def("ArrayTypeAnnotation")
 
 def("ObjectTypeAnnotation")
   .bases("Type")
-  .build("properties")
+  .build("properties")Al
```
gabelevi: Ugh stupid editor...

gabelevi:
Hey @benjamn, is this good to go?

benjamn:
Yep, thanks for the ping.

gabelevi:
Thanks a ton!

PR #93, opened 2015-03-02
XJS -> JSX

### LINKED ISSUE #91: XJS -> JSX
I wanted to give you a heads up about https://github.com/facebook/esprima/pull/85 (I know you're CC'ed but just in case) and figure out what we can do to make that transition smooth. The change should be easy to make (I literally just did a find+replace). I don't know how you want to version this lib - I'm sure we'll just bump major for esprima.

fkling:
Closes #91. I did a simple search/replace. I don't think there is anything else to be done here?

### DIFF
--- a/def/fb-harmony.js
+++ b/def/fb-harmony.js
@@ -7,67 +7,67 @@ var isString = builtin.string;
 var isBoolean = builtin.boolean;
 var defaults = require("../lib/shared").defaults;
 
-def("XJSAttribute")
+def("JSXAttribute")
     .bases("Node")
     .build("name", "value")
-    .field("name", or(def("XJSIdentifier"), def("XJSNamespacedName")))
+    .field("name", or(def("JSXIdentifier"), def("JSXNamespacedName")))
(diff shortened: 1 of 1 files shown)


fkling:
We could also have definitions for both, `XJS*` and `JSX*` elements. That would allow consumers to use older esprima-fb versions. But I don't know if this is really valuable.

benjamn:
Thanks! And yeah, I don't think anyone really depends on this part of the library.

sebmck:
@benjamn Any ETA on a release? Currently a blocker for babel/babel#930, thanks!

benjamn:
I think this has to be a minor version bump, just in case anyone was using the `XJS*` types. Is that ok?

fkling:
Sounds reasonable to me.

benjamn:
Ok, published to NPM as v0.7.0: http://npmjs.org/package/ast-types

sebmck:
Thanks @benjamn!

PR #95, opened 2015-03-06
Add `static` modifier to MethodDefinition

cpojer:


def/es6.js
```diff
@@ -78,11 +78,12 @@ def("PropertyPattern")
 
 def("MethodDefinition")
     .bases("Declaration")
-    .build("kind", "key", "value")
+    .build("kind", "key", "value", "static")
     .field("kind", or("init", "get", "set", ""))
-    .field("key", or(def("Literal"), def("Identifier"), def("Expression")))
+    .field("key", or(def("Literal"), def("Identifier")))
```
benjamn: The `.key` field can be an expression when `.computed` is true.

cpojer: uh oh, I think I copied different versions of this code block around and accidentally removed this. Will fix.

cpojer:
See https://github.com/benjamn/ast-types/pull/101

PR #99, opened 2015-03-22
StringLiteralTypeAnnotation should be based on Literal

cpojer:
I'm adding type annotation printing (wohoo) to recast and while I don't really know what StringLiteralTypeAnnotation does in flow, I think the ast-type for it should be based on Literal as well, just so that `nodeStr` in recast's printer module can be called on it.

cc @gabelevi

### DIFF
--- a/def/fb-harmony.js
+++ b/def/fb-harmony.js
@@ -119,6 +119,7 @@ def("StringTypeAnnotation")
 
 def("StringLiteralTypeAnnotation")
   .bases("Type")
+  .bases("Literal")
   .build("value", "raw")
   .field("value", isString)
   .field("raw", isString);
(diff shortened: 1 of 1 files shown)


cpojer:
+ ObjectTypeProperty should be based on Literal

gabelevi:
You can do things like

    var x: "hello" = "hello";
    var y: string = y;

So a string literal type annotation is a subtype of the string type annotation. 

None of the type annotations are literals and can't be used where a literal is required so I'm not sure if this PR is a good idea. String literal type annotations might pretty print the same way (object type annotations don't), but they're not literals.

benjamn:
I have to agree with @gabelevi here, especially since `Literal` nodes are `Expression`s, and I really don't think type annotations should count as expressions. Let's just change [`nodeStr`](https://github.com/benjamn/recast/blob/c9f3cd2b39313ed6711fa0c746004d93fe5acb14/lib/printer.js#L1286) to take a string instead of a string-valued `Literal`.

cpojer:
Oh I guess that is where my understanding of JS ASTs was not complete. I fixed the code in recast to remove the assertion, so we don't need this :)

Issue #94, opened 2015-03-04
Idea: "Virtual" type definitions


fkling:
The idea is to be able to treat different node types as if they were the same type. You already have a type which could be considered virtual: `Function`, which defines the shared properties of `FunctionExpression`s and `FunctionDeclaration`s. However, `FunctionExpression` and `FunctionDeclaration` have to explicitly "inherit" from `Function`.

Lets consider another example: `Identifier` and `JSXIdentifier`. For static analysis and code mods, I might want to treat both of these types as identical.

I would like to be able to write logic to process nodes independently of the concrete node type. Virtual types could be created dynamically and concrete node types could be associated with virtual types dynamically.

I'm not sure if what I'm saying makes sense at all , I'm not especially good at expressing my ideas. Or maybe something like this is already possible to do and I just don't know how. Happy to talk about this via a different channel.

benjamn:
I don't want to ignore the broader possibility, but in that particular case, I think `JSXIdentifier` should just be a subtype of `Identifier`, and then you could visit both of them with
```js
require("ast-types").visit(ast, {
  visitIdentifier: function(path) {
    if (path.value.type === "JSXIdentifier") {
      ...
    } else {
      ...
    }
    this.traverse(path);
  }
});
```

PR #100, opened 2015-03-23
Make JSXIdentifier an Identifier (not only a Node)

fkling:
From #94.

### DIFF
--- a/def/fb-harmony.js
+++ b/def/fb-harmony.js
@@ -18,7 +18,7 @@ def("JSXAttribute")
     ), defaults["null"]);
 
 def("JSXIdentifier")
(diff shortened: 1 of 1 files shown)


def/fb-harmony.js
```diff
@@ -150,7 +150,7 @@ def("FunctionTypeParam")
   .field("name", def("Identifier"))
   .field("typeAnnotation", def("Type"))
   .field("optional", isBoolean);
-
```
fkling: Apparently my editor stripped trailing white spaces. I can revert that if you want.

benjamn: All good!

PR #112, opened 2015-04-30
Compatibility with estree (specifically, babel-core/acorn)

tobich:
Adds definitions of ES6 modules according to the [estree](https://github.com/estree/estree) standard (as implemented by the babel-core/acorn parser.)


benjamn:
Just pushed b49b00811e65d0db55468ee2fae1111abaec5bc6 to master, which might allow you to rebase and get tests to pass without your latest two commits.

tobich:
"npm install -g npm@'>=1.4.3'" in .travis.yml is still required for the tests to pass (nodejs 0.8)

ai:
Any updates? This error makes a problem with Babel and Regenrator https://github.com/babel/babel/issues/1471

def/es6-modules.js
```diff
@@ -0,0 +1,80 @@
...
+
+// Based on https://github.com/estree/estree
+
+
+def("ModuleSpecifier2")
+    .bases("Node")
+    .build("local")
+    .field("local", def("Identifier"));
+
+def("ModuleSource");
```
benjamn: Where is this used/needed?

def/es6-modules.js
```diff
@@ -0,0 +1,80 @@
...
+var defaults = require("../lib/shared").defaults;
+var builtin = types.builtInTypes;
+var isString = builtin.string;
+
+// Based on https://github.com/estree/estree
+
+
+def("ModuleSpecifier2")
+    .bases("Node")
+    .build("local")
```
benjamn: Does this need to be buildable? If not, you can just remove the `.build("local")` line to make the type "abstract" (i.e. other types can inherit from it, but there won't be a builder called `builders.moduleSpecifier2`).

def/es6-modules.js
```diff
@@ -0,0 +1,80 @@
...
+
+def("ModuleSource");
+
+def("ImportSpecifier")
+    .bases("ModuleSpecifier2")
+    .build("imported")
+    .field("imported", def("Identifier"));
+
+def("ImportDefaultSpecifier")
+    .bases("ModuleSpecifier2")
```
benjamn: Semicolons please.

benjamn:
I plan to merge these `def`s into https://github.com/benjamn/ast-types/blob/master/def/es6.js the next time I make a breaking version change.

PR #118, opened 2015-07-16
Fix ClassProperty fb-harmony definition

gabelevi:
The key can also be a literal or an expression. Also, I don't see a good reason to have static in the build (not a big deal, can revert that part if you disagree)

### DIFF
--- a/def/fb-harmony.js
+++ b/def/fb-harmony.js
@@ -262,8 +262,7 @@ def("Function")
          defaults["null"]);
 
 def("ClassProperty")
-  .build("key", "value", "typeAnnotation", "static")
-  .field("key", def("Identifier"))
+  .build("key", "value", "typeAnnotation")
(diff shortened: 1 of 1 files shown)


benjamn:
Wait, what happened to the `key` field?

gabelevi:
It's defined in [`es6.js`](https://github.com/benjamn/ast-types/blob/c4d0280add526ebe38fa9698e056c8fe3fe8c639/def/es6.js#L141). The `fb-harmony.js` file is extending that core definition. This PR fixes a bug where we were actually making the `ClassProperty` node's key type more limited.

benjamn:
Ah, right.

benjamn:
https://github.com/benjamn/ast-types/releases/tag/v0.8.4

gabelevi:
Thanks!

PR #119, opened 2015-07-30
Add RestElement to ArrayExpression, NewExpression and CallExpression.

cpojer:
Recast PR incoming too. Babel uses RestElement.

### DIFF
--- a/def/es6.js
+++ b/def/es6.js
@@ -108,7 +108,12 @@ def("SpreadElement")
     .field("argument", def("Expression"));
 
 def("ArrayExpression")
-    .field("elements", [or(def("Expression"), def("SpreadElement"), null)]);
+    .field("elements", [or(
+        def("Expression"),
+        def("SpreadElement"),
+        def("RestElement"),
+        null
+    )]);
 
 def("NewExpression")
     .field("arguments", [or(def("Expression"), def("SpreadElement"))]);
--- a/def/fb-harmony.js
+++ b/def/fb-harmony.js
@@ -262,7 +262,7 @@ def("Function")
          defaults["null"]);
 
 def("ClassProperty")
-  .build("key", "value", "typeAnnotation")
+  .build("key", "value", "typeAnnotation", "static")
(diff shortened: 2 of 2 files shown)


def/es6.js
```diff
@@ -108,13 +108,26 @@ def("SpreadElement")
...
+        def("SpreadElement"),
+        def("RestElement")
+    )]);
 
 def("CallExpression")
-    .field("arguments", [or(def("Expression"), def("SpreadElement"))]);
+    .field("arguments", [or(
+        def("Expression"),
+        def("SpreadElement"),
+        def("RestElement")
```
benjamn: Can you give an example of a `CallExpression` that parses with a `RestElement`? My version of Babel/Acorn uses `SpreadElement`.

def/es6.js
```diff
@@ -108,13 +108,26 @@ def("SpreadElement")
...
+        null
+    )]);
 
 def("NewExpression")
-    .field("arguments", [or(def("Expression"), def("SpreadElement"))]);
+    .field("arguments", [or(
+        def("Expression"),
+        def("SpreadElement"),
+        def("RestElement")
+    )]);
```
benjamn: Can you give an example of a `NewExpression` that parses with a `RestElement`? My version of Babel/Acorn uses `SpreadElement`.

def/es6.js
```diff
@@ -108,13 +108,26 @@ def("SpreadElement")
     .field("argument", def("Expression"));
 
 def("ArrayExpression")
-    .field("elements", [or(def("Expression"), def("SpreadElement"), null)]);
+    .field("elements", [or(
+        def("Expression"),
+        def("SpreadElement"),
+        def("RestElement"),
+        null
+    )]);
```
benjamn: I think you'll only find a `RestElement` in an `ArrayPattern` (that is, an array expression on the left-hand side of an assignment or variable declaration).

cpojer:
Oh yes, you are right of course! Updated.

PR #120, opened 2015-07-30
Fix comments. Fixes recast/203

cpojer:
I originally mapped `CommentBlock` to `Block` etc. in babel.js but then Ben suggested to do the opposite thing.

### DIFF
--- a/def/core.js
+++ b/def/core.js
@@ -358,14 +358,10 @@ def("Comment")
     .field("leading", Boolean, defaults["true"])
     .field("trailing", Boolean, defaults["false"]);
 
-// Block comment. The .type really should be BlockComment rather than
-// Block, but that's what we're stuck with for now.
-def("Block")
+def("CommentBlock")
     .bases("Comment")
     .build("value", /*optional:*/ "leading", "trailing");
 
-// Single line comment. The .type really should be LineComment rather than
-// Line, but that's what we're stuck with for now.
-def("Line")
+def("CommentLine")
     .bases("Comment")
     .build("value", /*optional:*/ "leading", "trailing");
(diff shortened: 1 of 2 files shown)


def/core.js
```diff
@@ -360,12 +360,12 @@ def("Comment")
 
 // Block comment. The .type really should be BlockComment rather than
 // Block, but that's what we're stuck with for now.
-def("Block")
+def("CommentBlock")
     .bases("Comment")
     .build("value", /*optional:*/ "leading", "trailing");
```
benjamn: Oh, yeah, maybe `CommentBlock` and `CommentLine` should go in `def/babel.js`.

The order of evaluation doesn't matter, but it's nice to keep it clear that these types are Babel/Esprima-specific.

Also, you can delete the `//` comments above the definitions where I complain about the Esprima type names :)

cpojer: that's what I had before. But then Block and Line will still be in core.js?

benjamn: No I mean `Block` and `Line` would move to `def/esprima.js`, since that's where they always probably should have been. The abstract `Comment` supertype can remain in `def/core.js`, though.

cpojer:
@benjamn is this what you meant or would you prefer me to move CommentBlock and CommentLine into babel.js?

benjamn:
Yeah, sorry, I would prefer to put `CommentLine` and `ComentBlock` in `def/babel.js`.

I'm planning to make it easier to instantiate the type system with only one of `def/{esprima,babel}.js` sometime soon, so the distinction will become more meaningful.

Thanks for removing the actual comments.

cpojer:
But then you'll have to require babel.js for esprima.js to work? I don't think that makes sense unless you want me to copy the entire definition instead of just the alias.

benjamn:
Ah, you're right! I'd prefer not to copy the definitions. The current version looks good to me. Thanks for your patience.

benjamn:
After all that, this was necessary to fix a test in Recast that tried to call `recast.types.builders.line`: https://github.com/benjamn/ast-types/commit/fb7f78631bc417af9d08175dd4440712c7570065

PR #121, opened 2015-08-11
expose parameter metadata for builder functions

### LINKED ISSUE #84: Find out params order for particular builder
I want to automate building AST subtrees, something like:

```js
var types = require('ast-types');
var nodeParams = {...};
types.builders[nodeParams.type].apply(this, ???);
```

Is there a way to find out what is the proper order of params I should pass to a node builder of specific type?

There are [`.buildParams`](https://github.com/benjamn/ast-types/blob/27af2dff60cb27452b6ed43cf4fa295d8de471ea/lib/types.js#L462) defined on a type, but for now it seems impossible to get access to that list.
It would be nice to have something like `types.namedTypes[type].fields` I suppose?
Or to make `getFieldNames(node)` return properly ordered fields?
Or just to make any builder accept object with params, not only list of arguments.

I can probably do a PR, if you wish.

jamestalmage:
Closes #84 

```js
var types = require('ast-types');
var assert = require('assert');
var b = types.builders;
var n = types.namedTypes;

assert.strictEqual(b.whileStatement.paramCount, 2);

assert.deepEqual(b.whileStatement[0], {
  name: 'test',
  body: n.Expression
});

assert.deepEqual(b.whileStatement[1], {
  name: 'body',
  body: n.Statement
});
```

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -723,6 +723,31 @@ Dp.finalize = function() {
         if (self.buildable && self.supertypeList.lastIndexOf("Expression") >= 0) {
             wrapExpressionBuilderWithStatement(self.typeName);
         }
+
+        if (self.buildable) {
+            var buildFn = builders[getBuilderName(self.typeName)];
+
+            self.buildParams.forEach(function(param, i) {
+                var paramDef = {};
+
+                Object.defineProperty(paramDef, "name", {
+                    enumerable: true,
+                    value: param
+                });
+
+                Object.defineProperty(paramDef, "type", {
+                    enumerable: true,
+                    value: self.allFields[param].type
+                });
+
+                Object.defineProperty(buildFn, "" + i, {
+                    enumerable: true,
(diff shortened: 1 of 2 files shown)


jamestalmage:
Note that the discussion in #84 suggested a `length` property to determine the number of parameters, but that is already defined (and not configurable) for Functions.

It would be possible to generate build functions with the correct [arity](https://en.wikipedia.org/wiki/Arity), so that the `length` property would accurately reflect the number of parameters.

Note: the `length` property will be [configurable in es6](http://www.ecma-international.org/ecma-262/6.0/#sec-function-instances-length), but [not in es5](http://www.ecma-international.org/ecma-262/5.1/#sec-15.3.3.2).

demoneaux:
:+1: to the idea, although I think a better way to do this would be to add a `parameters` array, which removes the need for a separate `paramCount` property as `parameters.length` can be used instead.

PR #124, opened 2015-08-13
Fix ImportDefaultSpecifier AST conflict

marcioj:
This fix the problem that I commented in https://github.com/benjamn/recast/issues/206#issuecomment-129266283. I just used the same idea from [babel's ImportDefaultSpecifier](https://github.com/benjamn/ast-types/blob/master/def/babel.js#L58-L60): uses `ModuleSpecifier` as base node and build the needed field, in esprima case the `id` field.

//cc @benjamn

### DIFF
--- a/def/esprima.js
+++ b/def/esprima.js
@@ -51,15 +51,13 @@ def("ImportSpecifier")
 
 // import <* as id> from ...;
 def("ImportNamespaceSpecifier")
-    .bases("Specifier")
-    .build("id")
-    .field("id", def("Identifier"));
+    .bases("ModuleSpecifier")
+    .build("id");
 
 // import <id> from ...;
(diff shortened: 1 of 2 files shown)


def/esprima.js
```diff
@@ -57,9 +57,8 @@ def("ImportNamespaceSpecifier")
```
benjamn: Can you do the same thing for `ImportNamespaceSpecifier`?

marcioj: Done

benjamn:
Thanks!

benjamn:
Published version 0.8.9 to NPM: https://github.com/benjamn/ast-types/releases/tag/v0.8.9

marcioj:
Glad to help!

PR #134, opened 2015-11-09
DeclareExportDeclaration

gabelevi:
Flow is adding `declare export` declarations, which are very similar to `export` declarations. Currently Flow outputs the esprima representation of `export` declarations, so the `declare export` declarations are modeled similarly. Hopefully we'll move this over to match estree and babel sometime soon.

@benjamn - is it a problem that this refers to nodes declared in `esprima.js`?


gabelevi:
Ping @benjamn

benjamn:
Those definitions should be moved from `esprima.js` to `es6.js`, honestly.

PR #135, opened 2015-11-24
Added default behaviour to MemberExpression and a spec

juliankrispel:
cc @benjamn this adds default behaviour to MemberExpression so that computed is set to true if property is a `Literal`, another `MemberExpression` or a `BinaryExpression`.

### DIFF
--- a/def/core.js
+++ b/def/core.js
@@ -305,7 +305,15 @@ def("MemberExpression")
     .build("object", "property", "computed")
     .field("object", def("Expression"))
     .field("property", or(def("Identifier"), def("Expression")))
-    .field("computed", Boolean, defaults["false"]);
+    .field("computed", Boolean, function(){
+        var type = this.property.type;
+        if (type === 'Literal' ||
+            type === 'MemberExpression' ||
+            type === 'BinaryExpression') {
+            return true;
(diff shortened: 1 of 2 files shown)


juliankrispel:
ping @benjamn

benjamn:
Sorry for the wait. Happy to merge this as-is, but do you have any thoughts on whether it might be better to `return true` iff `type !== "Identifier"`? That way you wouldn't have to list out the alternatives. It's definitely not perfect, since you can have computed identifier properties, like `foo[bar]`. Then again, if there was a perfect test, we wouldn't need the third argument at all!

juliankrispel:
@benjamn thanks for getting back to me so quickly. Tell me if I'm being a little pushy :).

The only reason I didn't use `type !== "Identifier"` was to be more explicit. Because that'd assume that I can think of every single possibility here, which since I'm new to the project I am not :D

Am I making sense? I might be overthinking things.

juliankrispel:
I'm cool with just merging the changes without a discussion if you wanna get on with life and leave this behind :) :) :) :heart: @benjamn

juliankrispel:
haha thanks @benjamn <3

Issue #57, opened 2014-09-05
Proposal: Make the project generalized


Naddiseo:
Currently, this project is geared towards ES, but since the interface is so nice for building asts, I'd like to use it in other parsing projects.
Would it be possible to make the project more generalized so that it doesn't load all the types in `def/*` unless requested?

benjamn:
I still think this is a great idea, by the way. Hope to work on it for the next minor version (0.8.0).

jamestalmage:
bump!
`0.9.0` maybe?

Personally, I think the core components should be split out into another library (`ast-types-base` for sake of argument).

I like the syntax, but I think you should pass actual modules (or a custom function) instead of strings. The problem with strings would be accurately resolving them (especially if you do split into a separate module):

```js
var types = require('ast-types-base')(
  require('ast-types/def/core'),
  require('./myCustomDefs'),
  function(types) {
   types.def('MyType')
     //...
  }
);
```

I do like the `fork` option discussed, and that seems appropriate for `ast-types`, but not `ast-types-base` as I'm proposing. Perhaps strings would make sense there for specifying some of the built in stuff.

This would be the equivalent, less verbose version of above:

```js
require('ast-types').fork(
  'core',
  require('./myCustomDefs'),
  function(types) {
   types.def('MyType')
     //...
  }
);
```

One problem with the `fork` method as currently being discussed would be with `browserify` builds. Browserify will pick up and package everything in `def/*`, even if you aren't using any of it and creating an entirely custom AST.

jamestalmage:
@benjamn - anything I can do to move this forward?

bnjmnt4n:
Should probably be closed since #145 was merged nearly 2.5 years (!!!) ago.

benjamn:
Agreed, though I should mention the move to TypeScript (#300) has complicated this story, since the generated TypeScript declarations are fundamentally less modular/dynamic than the forking system.

PR #137, opened 2016-01-06
Add TypeAlias to scope::lookup()/declares()

danez:
Currently when looking up an Identifier that points to and TypeAlias it is not found by scope::lookup() or declares().

This PR adds TypeAlias to as valid target to the scope.

Example:

```js
type MyType = { x: string };

var foo: MyType = { x: "bar" };
```
The Identifier "MyType" of the TypeAnnotation in the variable declaration can now be looked up with this PR.

I also though about duplicate targets with this, but flow does not allow a TypeAlias to be named the same as a variable in the same scope, so no problem from this side.

This is needed for reactjs/react-docgen#48 which uses recast, so after this gets merged, it would be nice if you could create a new version of ast-type, and then of recast with this new version.

I tried creating a test, but couldn't get esprima to understand flow syntax.

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -34,7 +34,8 @@ function Scope(path, parentScope) {
         isGlobal: { value: !parentScope, enumerable: true },
         depth: { value: depth },
         parent: { value: parentScope },
-        bindings: { value: {} }
+        bindings: { value: {} },
+        types: { value: {} },
     });
 }
 
@@ -65,6 +66,11 @@ Sp.didScan = false;
 Sp.declares = function(name) {
     this.scan();
     return hasOwn.call(this.bindings, name);
+};
+
+Sp.declaresType = function(name) {
+    this.scan();
+    return hasOwn.call(this.types, name);
 };
 
 Sp.declareTemporary = function(prefix) {
@@ -115,7 +121,7 @@ Sp.scan = function(force) {
             // Empty out this.bindings, just in cases.
             delete this.bindings[name];
         }
-        scanScope(this.path, this.bindings);
+        scanScope(this.path, this.bindings, this.types);
         this.didScan = true;
     }
 };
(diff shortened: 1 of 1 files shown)


lib/scope.js
```diff
@@ -164,6 +164,9 @@ function recursiveScanScope(path, bindings) {
 
         recursiveScanChild(path.get("body"), bindings);
 
+    } else if (namedTypes.TypeAlias.check(node)) {
```
benjamn: My only request would be to add a defensive check that `namedTypes.TypeAlias` is defined:
```js
} else if (namedTypes.TypeAlias &&
            namedTypes.TypeAlias.check(node)) {
```
I want to make it easier to load selective subsets of AST types (#57), and I would like this code to continue working even if the Flow types are not loaded.

danez:
Ok I've added the check.

benjamn:
One more thing: I think it's worth commenting that while the type name is not strictly a variable, it does occupy the same logical namespace because it's not allowed to conflict with actual variable names.

To do this right, I think the `Scope` object would need to have a `.types` field in addition to the `.bindings` field, and then we could maybe also add `Scope.prototype.{declares,lookup}Type` methods. Not sure if you have time to look into that, but I wanted to make a note of it here so I don't forget.

danez:
I can do that if you want, but not sure if thats not to specific for flow, but it would probably also work for typescript. 
So creating a separate collection for types and own functions for lookup and declare? That should be fairly straight forward.

danez:
Ok I added everything as described. I did not split up the scan functions to not duplicate code.

benjamn:
[`recast@0.10.41`](https://github.com/benjamn/recast/releases/tag/v0.10.41) uses [`ast-types@0.8.15`](https://github.com/benjamn/ast-types/releases/tag/v0.8.15), which includes these changes

danez:
Thank you very much. That was quick.

PR #141, opened 2016-02-03
Add support for NullLiteralTypeAnnotation

phpnode:
Right now ast-types will throw on this syntax, because NullLiteralTypeAnnotation is not defined.

```js
type Foo = string | null;
```

This PR fixes the issue.


phpnode:
Btw I don't think the test failures are related to this change.

benjamn:
I'll look into the test failures.

PR #142, opened 2016-02-12
Fix default of object type call property.

garetht:
This fixes the default value for the `static` field in `ObjectTypeCallProperty`.


benjamn:
Why the close?

garetht:
Tests are failing.

garetht:
Looks like tests are also failing on master. Reopening.

benjamn:
Good catch, thanks!

PR #145, opened 2016-04-20
Make forkable

jamestalmage:
My attempt at fixing #57.

```js
// all submodules now export a function that takes a `fork` instance
module.exports = function (fork) {

  // this use of fork.use replaces "require" everywhere.
  var types = fork.use(require('../lib/types')); 
  
  // modules that exported something should now return it from the function instead.
  return exportsObject;
};
```

Fully customizable definitions can now be achieved:

```js
var types = require('ast-types/fork')([
  require('ast-types/def/core'); // just the core
]);
```

One potential downside (and I don't see it being a big deal):

```js
var types1 = require('ast-types/fork')([
  require('ast-types/def/core');
]);

var types2 = require('ast-types/fork')([
  require('ast-types/def/core');
]);

// This throws with "[object Object] is not a Def"
// It probably should throw, but the error message needs improvement.
types1.Type.def('Node').isSupertypeOf(types2.Type.def('Expression'))
```

### DIFF
--- a/def/babel.js
+++ b/def/babel.js
@@ -1,9 +1,7 @@
 module.exports = function (fork) {
     fork.use(require("./es7"));
 
-    // var types = fork.types;
     var types = fork.use(require("../lib/types"));
-    // var defaults = fork.shared.defaults;
     var defaults = fork.use(require("../lib/shared")).defaults;
     var def = types.Type.def;
     var or = types.Type.or;
--- a/def/babel.js
+++ b/def/babel.js
@@ -1,105 +1,109 @@
-require("./es7");
+module.exports = function (fork) {
+    fork.use(require("./es7"));
 
-var types = require("../lib/types");
-var defaults = require("../lib/shared").defaults;
-var def = types.Type.def;
-var or = types.Type.or;
+    // var types = fork.types;
+    var types = fork.use(require("../lib/types"));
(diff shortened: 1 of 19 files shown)


benjamn:
Thanks for this @jamestalmage!

If we're going to change all that whitespace anyway, I would really love to finally switch to using two spaces per tab for indentation. Would that be easy to fix?

jamestalmage:
I would think so. I will update soon.

hzoo:
@jamestalmage Can you update with what needs to be done here?

jamestalmage:
Rebased, and eliminated some irrelevant style-only changes that were cluttering the diff:

Best viewed via https://github.com/benjamn/ast-types/pull/145/files?w=1

@benjamn - I never did change to two tab spaces, but now I am concerned doing so would create headaches for @hzoo rebasing #162 once this is merged.

Would you prefer I hold off on the whitespace changes for now? @hzoo?

hzoo:
Might make it easier to read but you can always use `?w=1` and in my case i'm adding a new file so it's fine

benjamn:
Thanks again @jamestalmage, especially for your patience. 🍴 🎉

jamestalmage:
:tada: happy this made it in!

PR #156, opened 2016-06-23
Pick up indexers correctly in ObjectTypeAnnotation (flow)

keyanzhang:
Seems like the `field`s were added but the `build` spec hasn't been updated. Thanks!


benjamn:
I don't always add `.build` params for fields that don't seem like they would be commonly used, but I'll defer to anyone who tells me otherwise :)

keyanzhang:
Thanks!

PR #157, opened 2016-06-24
[do not merge] scope refactor/cleanup + discuss

bmeck:
Leaving this here for discussion purposes while I start work. These are somewhat breaking changes that I will be going over the reason why on each.

Plan:
1. have "kind" on scopes: Global, Object (`with` like), Block, Function, Catch, Module corresponding to EnvironmentRecords in spec
   
   This aids in some features like `let`/`const` being supported, and would avoid the strange add/delete of binding for Catch scopes. New path would be for code to have knowledge of what scope types variables can bind to.
   
   | kind | types |
   | --- | --- |
   | var | Function,Global,Module |
   | let | Block,Function,Global,Module |
   | const | Block,Function,Global,Module |
   | params | Function |
   | catch | Catch |
   | import | Module |
   
   This would change the scanning code since it would scan all the children that could affect the parent scope, namely this means only scopes that can hold all kinds of variables would stop a scan (Function,Global, and Module)
2. Setup Global scope in `.visit(ast, visitors, {globalScope:})`
   
   this would mean that you could reuse global scopes between visits and share a global scope between ASTs.
3. Upgrade to newer JS, supported by Node v4+
4. JSDoc

### DIFF
--- a/lib/equiv.js
+++ b/lib/equiv.js
@@ -1,3 +1,4 @@
+'use strict';
 var types = require("../main");
 var getFieldNames = types.getFieldNames;
 var getFieldValue = types.getFieldValue;
--- a/lib/node-path.js
+++ b/lib/node-path.js
@@ -1,3 +1,4 @@
+'use strict';
 var types = require("./types");
 var n = types.namedTypes;
 var b = types.builders;
--- a/lib/path-visitor.js
+++ b/lib/path-visitor.js
@@ -1,3 +1,4 @@
+'use strict';
 var types = require("./types");
 var NodePath = require("./node-path");
 var Printable = types.namedTypes.Printable;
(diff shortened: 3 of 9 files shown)


jedwards1211:
@bmeck this is similar to my open PR for type parameter scopes.
By `params` in the table do you mean the arguments to a function or type parameters like `function foo<T>(): T { }`?
In `ClassDeclaration`s, `ClassExpression`s, `InterfaceDeclaration`s, and `TypeAlias`es with `TypeParameter`s, the type parameter scope is confined to the declaration, so a special kind of scope has to be created for those too.
(This is all referring to Flow, I don't know 100% how much this applies to TypeScript but I imagine it's similar)
Should I pick this back up?  I have a need for an accurate system to find undeclared identifiers.

PR #159, opened 2016-06-28
Add TypeParameter and NullTypeAnnotation (flow)

keyanzhang:
This PR adds support for 2 types that the flow parser emits.

An example can be found here: https://astexplorer.net/#/KAEq1HWZBo

Thanks!

### DIFF
--- a/def/flow.js
+++ b/def/flow.js
@@ -183,7 +183,7 @@ def("TypeParameter")
          defaults["null"])
   .field("bound",
          or(def("TypeAnnotation"), null),
-         defaults["null"])
+         defaults["null"]);
 
 def("Function")
   .field("returnType",
--- a/def/flow.js
+++ b/def/flow.js
@@ -168,7 +168,7 @@ def("Identifier")
 def("TypeParameterDeclaration")
   .bases("Node")
   .build("params")
-  .field("params", [or(def("Identifier"), def("TypeParameter"))]);
+  .field("params", [def("TypeParameter")]);
 
 def("TypeParameterInstantiation")
   .bases("Node")
@@ -176,8 +176,9 @@ def("TypeParameterInstantiation")
   .field("params", [def("Type")]);
 
 def("TypeParameter")
-  .bases("Identifier")
-  .build("variance", "bound")
+  .bases("Type")
+  .build("name", "variance", "bound")
+  .field("name", String)
   .field("variance",
          or("plus", "minus", null),
          defaults["null"])
--- a/def/flow.js
+++ b/def/flow.js
@@ -64,6 +64,10 @@ def("NullLiteralTypeAnnotation")
   .bases("Type")
   .build();
 
+def("NullTypeAnnotation")
+  .bases("Type")
+  .build();
+
 def("ThisTypeAnnotation")
   .bases("Type")
   .build();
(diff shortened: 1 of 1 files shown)


def/flow.js
```diff
@@ -164,13 +168,23 @@ def("Identifier")
...
-  .field("params", [def("Identifier")]);
+  .field("params", [or(def("Identifier"), def("TypeParameter"))]);
 
 def("TypeParameterInstantiation")
   .bases("Node")
   .build("params")
   .field("params", [def("Type")]);
 
+def("TypeParameter")
+  .bases("Identifier")
```
samwgoldman: In Flow, this node is not a subtype of Identifier, but is its own node with a name, location, variance, and bound.

def/flow.js
```diff
@@ -164,13 +168,23 @@ def("Identifier")
 def("TypeParameterDeclaration")
   .bases("Node")
   .build("params")
-  .field("params", [def("Identifier")]);
+  .field("params", [or(def("Identifier"), def("TypeParameter"))]);
```
samwgoldman: There should only be TypeParameter nodes in here.

def/flow.js
keyanzhang: Cool, so its `bases` should just be `"Type"` then?

samwgoldman:
Thanks for doing this! The TypeParameter stuff seems a little off to me, but I'm actually not sure the entire ecosystem agrees on this representation. The `flow ast` output is canonical, I think. For Babel we hacked some stuff onto Identifier for backwards compat, rather than creating a new node type, but Babel should deprecate the use of Identifier in this position and move to TypeParameter if it hasn't already.

def/flow.js
samwgoldman: Yeah, I think so. `"Identifier"` from core would be fine, but since flow augments it with `typeParameter`—which `TypeParameter` doesn't have—I think you need to base it on `Type` and add a `"name"` field.

That said, this is the first time I've looked at this repo, so there might be a better way.

keyanzhang:
@samwgoldman jscodeshift now supports ["choosing your parser"](https://github.com/facebook/jscodeshift/releases/tag/v0.3.21) so we've been using the Flow parser directly. I think babel 5 uses `Identifier` but babel 6 has already been using `TypeParameter`. If you switch the parser to babylon6 [here](https://astexplorer.net/#/KAEq1HWZBo) you'll see that the `TypeParameter` part is the same.

def/flow.js
benjamn: This sounds right to me (that is, `.bases("Type").field("name", String)`).

keyanzhang: Fun fact: I used `Type` first but switched to `Identifier` because I saw [this](https://github.com/facebook/flow/blob/master/src/parser/test/custom_ast_types.js#L6-L10) :)

samwgoldman:
That's great—I think if Flow and Babylon6 agree, then that's the way to go here, too.

def/flow.js
keyanzhang: Updated per your feedback. Thanks!

samwgoldman:
Looks good to me! Thanks for doing this, Keyan :)

def/flow.js
```diff
@@ -164,13 +168,24 @@ def("Identifier")
...
   .bases("Node")
   .build("params")
   .field("params", [def("Type")]);
 
+def("TypeParameter")
+  .bases("Type")
+  .build("name", "variance", "bound")
+  .field("name", String)
+  .field("variance",
+         or("plus", "minus", null),
```
benjamn: Kind of odd this isn't just `+` and `-`…

samwgoldman: Yeah, I can see that. Internally in Flow we use a variant type `Plus | Minus`, but that we opted to convert that to "plus" and "minus" instead of "+" and "-" in our JS AST is an accident of history.

PR #160, opened 2016-06-29
fix: fixes error in scope traversal

KnisterPeter:
If childPath.value and child are both 'default' array they are not of
the same instance, but we must not throw an error in this case.


benjamn:
Thanks @KnisterPeter!

Issue #132, opened 2015-11-03
Support for new babel types


tmcw:
Re: https://github.com/babel/babel/issues/2769

Babel 6.0.0 introduces a new NumericLiteral type that currently is not present in ast-types and breaks the tree traversal.

benjamn:
Have time to submit a pull request?

hzoo:
It's more than just `NumericLiteral` - https://github.com/babel/babel/blob/master/doc/ast/spec.md, https://github.com/babel/babel/blob/master/CHANGELOG.md#600

Example changes - https://github.com/babel/acorn-to-esprima/pull/3/files#diff-d45f2be38f06b237a94ba1f97798af79R53

benjamn:
I'm guessing @tmcw doesn't have time for a pull request, which is totally fine. I'll do this as soon as I can.

hzoo:
Shouldn't be too bad. From the changelog

Rename RegexLiteral node to RegExpLiteral.
Rename NumberLiteral node to NumericLiteral.

The MethodDefinition node type has been renamed to ClassMethod and it's FunctionExpression value property has been coerced into the main method node.
The Property node type has been renamed to ObjectProperty.
The Property node type with the boolean flag method has been renamed to ObjectMethod and it's FunctionExpression value property has been coerced into the main method node.
The Literal node type has been unoverloaded and split into BooleanLiteral, RegExpLiteral, NumericLiteral, StringLiteral and NullLiteral.
The SpreadProperty (from object-rest-spread) node type has been split into RestProperty (for ObjectPattern) and SpreadProperty (for ObjectExpression)

benjamn:
Those are all really worthwhile changes, by the way.

hzoo:
Should probably propose them to ESTree anyway (lgtm)

jamestalmage:
I think this is going to be difficult w/o #57.

For example: attempting to add RegExpLiteral [per the spec](https://github.com/babel/babel/blob/master/doc/ast/spec.md#literals), I added the following definition:

```js
def("RegExpLiteral")
    .bases("Literal")
    .build("pattern", "flags")
    .field("pattern", String)
    .field("flags", String);
```

Trying to build a `RegExpLiteral` Node, you get the following error:

```js
b.regExpLiteral("my-pattern", "ig");
```
```
 Error: no value or default function given for field "value" of RegExpLiteral("pattern": string, "flags": string)
```

[...]

PR #162, opened 2016-07-04
Babel 6 types

hzoo:
Fixes https://github.com/benjamn/ast-types/issues/132
For babel pr: https://github.com/babel/babel/pull/3561
Also ref https://github.com/benjamn/recast/pull/298

First 3 commits are from https://github.com/benjamn/ast-types/pull/145 (fixed merge conflicts by preferring the changes in the PR).

You can review that PR or this one with whitespace off for a better diff https://github.com/benjamn/ast-types/pull/162/files?w=1.

My commit adds the babel 6 types (not sure if it's right though, so please review!).

References:
- babel-types definitions: https://github.com/babel/babel/tree/master/packages/babel-types/src/definitions
- babylon ast differences https://github.com/babel/babylon#output

cc @jamestalmage @benjamn

### DIFF
--- a/def/babel6.js
+++ b/def/babel6.js
@@ -1,0 +1,134 @@
+module.exports = function (fork) {
+    fork.use(require("./babel"));
+    fork.use(require("./flow"));
+
+    // var types = fork.types;
+    var types = fork.use(require("../lib/types"));
+    // var defaults = fork.shared.defaults;
+    var defaults = fork.use(require("../lib/shared")).defaults;
+    var def = types.Type.def;
+    var or = types.Type.or;
+
+    def("Directive")
+        .bases("Node")
+        .build("value")
+        .field("value", def("DirectiveLiteral"));
+
+    def("DirectiveLiteral")
+        .bases("Node", "Expression")
+        .build("value")
+        .field("value", String, defaults["use strict"]);
+
+    def("BlockStatement")
+        .bases("Statement")
+        .build("body")
+        .field("body", [def("Statement")])
+        .field("directives", [def("Directive")], defaults.emptyArray);
+
+    def("Program")
+        .bases("Node")
+        .build("body")
+        .field("body", [def("Statement")])
+        .field("directives", [def("Directive")], defaults.emptyArray);
+
+    // Split Literal
+    def("StringLiteral")
+        .bases("Literal")
+        .build("value")
+        .field("value", String);
+
+    def("NumericLiteral")
+        .bases("Literal")
+        .build("value")
+        .field("value", Number);
+
+    def("NullLiteral")
+        .bases("Literal")
+        .build();
+
+    def("BooleanLiteral")
+        .bases("Literal")
+        .build("value")
+        .field("value", Boolean);
+
+    def("RegExpLiteral")
+        .bases("Literal")
+        .build("pattern", "flags")
+        .field("pattern", String)
+        .field("flags", String);
+
+    // Split Property -> ObjectProperty and ObjectMethod
+    def("ObjectExpression")
+        .bases("Expression")
+        .build("properties")
+        .field("properties", [or(def("Property"), def("ObjectMethod"), def("ObjectProperty"), def("SpreadProperty"))]);
+
+    // ObjectMethod hoist .value properties to own properties
+    def("ObjectMethod")
+        .bases("Node", "Function")
+        .build("kind", "key", "params", "body", "computed")
+        .field("kind", or("method", "get", "set"))
+        .field("key", or(def("Literal"), def("Identifier"), def("Expression")))
+        .field("params", [def("Pattern")])
+        .field("body", def("BlockStatement"))
+        .field("computed", Boolean, defaults["false"])
+        .field("generator", Boolean, defaults["false"])
+        .field("async", Boolean, defaults["false"])
+        .field("decorators",
+            or([def("Decorator")], null),
+            defaults["null"]);
+
+    def("ObjectProperty")
+        .bases("Node")
+        .build("key", "value")
+        .field("key", or(def("Literal"), def("Identifier"), def("Expression")))
+        .field("value", or(def("Expression"), def("Pattern")))
+        .field("computed", Boolean, defaults["false"]);
+
+    // MethodDefinition -> ClassMethod
+    def("ClassBody")
+        .bases("Declaration")
+        .build("body")
+        .field("body", [or(def("ClassMethod"), def("ClassProperty"))]);
+
+    def("ClassMethod")
+        .bases("Declaration", "Function")
+        .build("kind", "key", "params", "body", "computed", "static")
+        .field("kind", or("get", "set", "method", "constructor"))
+        .field("key", or(def("Literal"), def("Identifier"), def("Expression")))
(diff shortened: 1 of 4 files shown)


def/babel6.js
```diff
@@ -0,0 +1,117 @@
+module.exports = function (fork) {
+    fork.use(require("./babel"));
+    fork.use(require("./flow"));
```
hzoo: Do I need to fork both? or should babel fork flow instead of es7?

def/babel6.js
```diff
@@ -0,0 +1,117 @@
...
+
+    def("DirectiveLiteral")
+        .build("value")
+        .field("value", String, defaults["use strict"]);
+
+    def("BlockStatement")
+        .bases("Statement")
+        .build("body")
+        .field("body", [def("Statement")])
+        .field("directives", [def("Directive")]); // add directive
```
hzoo: Should this only have `.field("directives", [def("Directive")]);` or include the bases/build even though its already defined?

hzoo:
General question - is it better to just define a new type with everything from scratch (base, build, field)? Some of the changes are literally just a name change so that probably doesn't have to be.

hzoo:
Should I be modifying main.js?

def/babel6.js
jamestalmage: I think you should fork both. It's conceivable people will want to use the types defined in `./babel` without those defined in `./flow`, right? If `./babel` forks `./flow`, it should imply that those type definitions somehow depend on `./flow`.

def/babel6.js
jamestalmage: Does this change the arguments used to build? If so, you need to redefine the `.build`. Otherwise, I don't think you need to redefine them.

jamestalmage:
> fixed merge conflicts by preferring the changes in the PR

That was so long ago, there are probably changes that should have been merged more carefully. I can take a look at #145 again tonight.

def/babel6.js
hzoo: I guess it just depends on whether you want directives to be in the the builder or not?

jamestalmage: I would suggest that you try to make the builders defined for Babel compatibility as similar to the `babel-types` builders as possible. Trying to context switch between `babel-types` / `ast-types` builders is already enough to drive you crazy, no need make `ast-types/babel` some weird mixture of the two and further confuse things.

jamestalmage:
> General question - is it better to just define a new type with everything from scratch (base, build, field)? Some of the changes are literally just a name change so that probably doesn't have to be.

My gut says just redefine. Babel has moved far enough away from esprima compatibility, that attempting to share base types seems likely to create some amount of pain, and I don't really see much of an upside. We are to the point of needing a custom fork anyways, so might as well just make the entire tree conform to the Babel way of doing things.

That said, simple name changes can probably just be accomplished with:

```js
def('newName').bases('oldName')
```

def/babel6.js
hzoo: Yes it has been a very long few days and pretty frustrating

hzoo:
Rebased again

def/babel6.js
```diff
@@ -10,47 +10,49 @@ module.exports = function (fork) {
...
         .bases("Node")
         .build("body")
         .field("body", [def("Statement")])
-        .field("directives", [def("Directive")]); // add directive
+        .field("directives", [def("Directive")], defaults.emptyArray);
 
     // Split Literal
     def("StringLiteral")
-        .bases("Literal")
+        .bases("Node", "Expression")
```
benjamn: `Node` is not important here, nor for the other Babel `*Literal` types, because `Expression <: Node`.

benjamn:
1. Can you rebase now that #145 has been merged?
2. I believe you can solve the test failures by creating multiple forks for different tests, instead of using the default fork that includes all the defs.

def/babel6.js
```diff
@@ -0,0 +1,125 @@
...
+        .field("async", Boolean, defaults["false"])
+        .field("decorators",
+            or([def("Decorator")], null),
+            defaults["null"]);
+
+    // Split into RestProperty and SpreadProperty
+    def("ObjectPattern")
+        .bases("Pattern")
+        .build("properties")
+        .field("properties", [or(def("RestProperty"), def("ObjectProperty"))])
```
mroch: it should be `Property`, not `ObjectProperty`, per https://github.com/estree/estree/blob/f32646033ed12e6aaa7f5cd496d936a8f8de194e/es2015.md#objectpattern

There are a couple of places in ESTree where nodes have different `type` values like this, but I don't think ast-types has a way to represent this.

suchipi:
I'd like to see this merged... how can I contribute? What needs to be done to fix the tests? I'm not familiar enough with the forking model in this codebase to understand how to fix failures.

def/babel6.js
hzoo: What do you mean here though? I'm defining all the different types from ESTree (including ObjectProperty above) - I thought we would be forking to support the different AST types in babel (6).

https://github.com/babel/babylon#output

mroch: Note how ObjectPattern.properties is an array of ObjectProperty's, like you have, but ObjectProperty's type field is "Property", not "ObjectProperty." def("ObjectProperty") requires an ast with { type: "ObjectProperty" }, but estree says it should be { type: "Property" }

hzoo: After v6, Babel isn't following ESTree completely so I had to make this PR to make it work with recast. Yeah it's totally correct it should be Property but Babylon won't produce a type of Property anymore so I'm using ObjectProperty if that makes sense (unless I'm misunderstanding something)

hzoo:
Looked through https://github.com/babel/babel/commits/master/packages/babel-generator to find new syntax.

def/babel6.js
```diff
@@ -0,0 +1,134 @@
...
+      .bases("Node")
+      .build("argument")
+      .field("argument", def("Expression"));
+
+    def("RestProperty")
+      .bases("Node")
+      .build("argument")
+      .field("argument", def("Expression"));
+
+    def("ForAwaitStatement")
```
hzoo: https://github.com/babel/babel/commit/26e79c543374e99e6742eac69d8f1f9d4ecf5ed1

def/flow.js
```diff
@@ -13,6 +13,10 @@ module.exports = function (fork) {
       .bases("Type")
       .build();
 
+    def("EmptyTypeAnnotation")
```
hzoo: https://github.com/babel/babel/pull/4724

def/flow.js
```diff
@@ -268,6 +281,11 @@ module.exports = function (fork) {
       .field("id", or(def("Identifier"), def("Literal")))
       .field("body", def("BlockStatement"));
 
+    def("DeclareModuleExports")
```
hzoo: https://github.com/babel/babel/pull/3570

hzoo:
Finally got around to fixing these.. @suchipi (I didn't understand the forking either).

Tested with https://github.com/benjamn/recast/pull/299 (would error that node isn't printable, etc)

hzoo:
Thanks @benjamn!!! much excite 😍

PR #169, opened 2016-08-06
Add DeclareModuleExports (Flow AST)

motiz88:
Encountered in flow-parser's test fixtures.
Definition basically lifted from [facebook/flow/.../custom_ast_types.js](https://github.com/facebook/flow/blob/master/src/parser/test/custom_ast_types.js).

### DIFF
--- a/def/flow.js
+++ b/def/flow.js
@@ -287,4 +287,9 @@ module.exports = function (fork) {
         def("Literal"),
         null
       ), defaults["null"]);
+
+    def("DeclareModuleExports")
+      .bases("Statement")
+      .build("typeAnnotation")
+      .field("typeAnnotation", def("TypeAnnotation"));
 };


def/flow.js
```diff
@@ -287,4 +287,9 @@ module.exports = function (fork) {
         def("Literal"),
         null
       ), defaults["null"]);
+
+    def("DeclareModuleExports")
+      .bases("Statement")
```
benjamn: Not `.bases("Declaration")`?

motiz88: I'm not sure. That's copied directly from Flow's source; also, the surface
syntax is `declare module.exports = foo` which looks kind of like an
assignment statement... :smile:

hzoo:
Added in https://github.com/benjamn/ast-types/pull/162/files#r86535999

benjamn:
Thanks @hzoo and @motiz88! Closing because this seems fixed already. Sorry it took that long!

PR #173, opened 2016-08-16
Provide missing importKind field definition on ImportDeclaration

jessebeach:
Imports can be inflected with a kind value:

``` javascript
import a from 'b';
import type c from 'b';
```

Whereas recast supports passing in a value for an `importKind` field on an `ImportDeclaration` node, ast-types lacks this field definition: https://github.com/benjamn/recast/blob/master/lib/printer.js#L480

I added the field and limited the allowed values to the strings "value" and "type". I don't know of any proposed analogs to "type", so for now it seems fine to just hardcode the two values instead of supporting a literal.

### DIFF
--- a/def/esprima.js
+++ b/def/esprima.js
@@ -90,7 +90,9 @@ module.exports = function (fork) {
       .field("importKind", or(
         "value",
         "type"
-      ), defaults["value"]);
+      ), function() {
+        return "value";
+      });
 
     def("Block")
       .bases("Comment")
--- a/def/esprima.js
+++ b/def/esprima.js
@@ -80,13 +80,17 @@ module.exports = function (fork) {
 
     def("ImportDeclaration")
       .bases("Declaration")
-      .build("specifiers", "source")
+      .build("specifiers", "source", "importKind")
       .field("specifiers", [or(
         def("ImportSpecifier"),
         def("ImportNamespaceSpecifier"),
         def("ImportDefaultSpecifier")
       )], defaults.emptyArray)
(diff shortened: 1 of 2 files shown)


def/esprima.js
```diff
@@ -80,13 +80,17 @@ module.exports = function (fork) {
...
         def("ImportSpecifier"),
         def("ImportNamespaceSpecifier"),
         def("ImportDefaultSpecifier")
       )], defaults.emptyArray)
-      .field("source", def("Literal"));
+      .field("source", def("Literal"))
+      .field("importKind", or(
+        "value",
+        "type"
+      ), defaults["value"]);
```
benjamn: The third argument to `.field` is supposed to be a function that returns a default value when called, so I think this should be `function () { return "value" }` instead of `defaults["value"]`.

jessebeach:
```
exports.defaults = {
    // Functions were used because (among other reasons) that's the most
    // elegant way to allow for the emptyArray one always to give a new
    // array instance.
    "null": function () { return null },
    "emptyArray": function () { return [] }, 
    "false": function () { return false },
    "true": function () { return true },
    "undefined": function () {}
};
```

Indeed, updating...

spudly:
Any chance this will get merged soon?

jessebeach:
Keeping it fresh on `benjamn:master`.

benjamn:
Thanks for this PR! And thanks for the reminder :)

jessebeach:
Thanks @benjamn!

PR #187, opened 2016-10-13
Add ForAwaitStatement

### LINKED ISSUE #186: ForAwaitStatement
The [async iteration proposal](https://github.com/tc39/proposal-async-iteration) is stage 3 and ready for implementation. Babylon adds support and uses a `ForAwaitStatement`, which is identical to the `ForOf` statement except for the name.

samwgoldman:
Fixes #186


hzoo:
Added this as part of https://github.com/benjamn/ast-types/pull/162 and i think we wouldn't put it in babel.js anymore since it's only for babel 6?

PR #194, opened 2016-11-08
allow older nodes in ClassBody body field

hzoo:
Ref https://github.com/reactjs/react-codemod/issues/88

Might need some other changes actually - need to check the overrides


hzoo:
This is what is needed? Every override requires allowing all the older nodes

PR #201, opened 2016-12-30
Add Flow's DeclareExportAllDeclaration type

jlongster:
Flow has a node type `DeclareExportAllDeclaration` that is generated from this code:

```js
declare export * from "foo.js"
```

This PR adds this node type.


benjamn:
Thanks!

PR #205, opened 2017-02-01
add Babel's NumericLiteralTypeAnnotation

danharper:
Flow's parser names it `NumberLiteralTypeAnnotation`, but Babylon currently names `NumericLiteralTypeAnnotation` ([it's being renamed for v7](https://github.com/babel/babylon/issues/329)).

Found this was missing when using `prettier`.


danharper:
Sure, I couldn't really decide as it's a Babel-specific Flow deviation. I'll make the change 👍

PR #209, opened 2017-03-18
Let `flow` be opt-in to babel6.

jdalton:
This makes flow types opt-in (since they're loaded in the main.js already).


jdalton:
If this is seen as too big a change I could create a babel6-no-flow.js file instead and have babel6.js and babel6-no-flow.js use a babel6-base.js or something (as I'm typing it I'm liking that more).

jdalton:
Updated to preserve the existing `babel6` which loads `babel` and `flow` but added `babel6-core` which is without `flow` 😄

jdalton:
Locally the unit tests pass. I'm not sure what's up with travis.

jdalton:
Raising the test bar didn't help travis either 😕

.travis.yml
```diff
@@ -1,21 +1,8 @@
...
 
 sudo: false
-  
-before_install:
-  npm install -g npm@'>=1.4.3'
-
-matrix:
-  allow_failures:
-    - node_js: "0.8"
-    - node_js: "0.6"
```
benjamn: Let's leave `.travis.yml` alone in this PR, especially since I just [updated it](https://github.com/benjamn/ast-types/blob/master/.travis.yml), so there may be rebase conflicts.

jdalton: Sure thing. Any idea what the fails where about?

benjamn: It was mostly dumb stuff like `require("babel-core").parse` no longer existing after the Greenkeeper update, and some unrelated changes to Babel 6 type definitions. As for `.travis.yml`, I just no longer care about Node pre-0.10 (those tests haven't been passing for a while, because of `^` version constraints in `package.json` files).

jdalton:
Rebased and tests pass 😄

PR #232, opened 2017-09-28
Fixes lack of support for ObjectTypeSpreadProperty in ObjectTypeProperty properties

0x24a537r9:
I think it was somehow missed when ObjectTypeSpreadProperty support was first added, which is odd because this is the only place ObjectTypeSpreadProperty could actually be used: http://astexplorer.net/#/gist/fd19e271d1c2ef14b2720f695ea01ce3/3b91af647e3bbfdb3d3ac1757ff16fbbc4a30ccd

### DIFF
--- a/def/flow.js
+++ b/def/flow.js
@@ -116,7 +116,7 @@ module.exports = function (fork) {
     def("ObjectTypeAnnotation")
       .bases("Type")
       .build("properties", "indexers", "callProperties")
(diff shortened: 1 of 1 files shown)


0x24a537r9:
@benjamn Should be a pretty quick review :)

0x24a537r9:
@benjamn Any way I could get a new npm publish with this change? That way I can update recast with this as mentioned in https://github.com/benjamn/recast/pull/431. Thanks!

benjamn:
@0x24a537r9 Done!

PR #235, opened 2017-10-06
Add function AssignmentPattern to function scope

baremetalfreak:
Default funtion arguments are compiled by bable into AssignmentPattern.
We need to add it into scope for lookup.


benjamn:
Thanks for catching this, @baremetalfreak! Sorry I took so long to merge this.

baremetalfreak:
My pleasure. Thank you for this great tool!

PR #241, opened 2017-12-12
`**` exponentiation operator has higher precedence than multiply/divide/modulo...

GerHobbelt:
..., as per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence

### DIFF
--- a/lib/node-path.js
+++ b/lib/node-path.js
@@ -308,7 +308,8 @@ module.exports = function (fork) {
     }
 
     var PRECEDENCE = {};
-    [["||"],
+    [
+        ["||"],
         ["&&"],
         ["|"],
         ["^"],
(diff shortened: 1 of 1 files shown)


bnjmnt4n:
Noticed while looking at ast-types and Recast that there is quite a bit of duplication between `node-path` and `fast-path`‘s `needParens`. Seeing that only Recast uses that for printing, is there a reason for this?

On topic now, Recast had a similar PR a couple of years back: https://github.com/benjamn/recast/pull/285, where the `**` operator was added with the same precedence as `*`, `/` and `%`. Not too sure about the semantics of these operations, so just putting this up for someone more well-versed to weigh in.

benjamn:
@bnjmnt4n A while back I put in some effort to minimize/eliminate `ast-types` dependencies for `recast.parse` and `recast.print` (though of course `recast.visit` and `recast.types` still depend on `ast-types`). Moving to `FastPath` and its implementation of `needsParens` was part of that work. `FastPath` is a less ergonomic abstraction than `NodePath`, but it's faster for `recast`'s needs because it keeps track of the current path using a mutable stack instead of lots of separate `NodePath` objects.

PR #243, opened 2018-01-03
Add JSXFragment to defs

yangshun:
Based on the updated [JSX AST spec](https://github.com/facebook/jsx/blob/master/AST.md#jsx-fragment)


yangshun:
@benjamn Would you mind taking a look at this?

benjamn:
@yangshun Looks good to me! Thanks for taking the time to put this PR together, and for validating it in production.

PR #245, opened 2018-01-27
Base `Noop` node on `Statement` to match `EmptyStatement` behavior.

eventualbuddha:
Recast treats them as effectively the same for printing purposes, so in order to prevent recast from crashing on `Noop` nodes when printing we need to treat them as statements. See https://github.com/square/babel-codemod/issues/43.

### DIFF
--- a/def/babel-core.js
+++ b/def/babel-core.js
@@ -7,7 +7,7 @@ module.exports = function (fork) {
   var or = types.Type.or;
 
   def("Noop")
-    .bases("Node")
+    .bases("Statement")
     .build();
 
   def("DoExpression")


eventualbuddha:
It may be that this fix is not what we actually want and recast should be updated to handle printing `Noop` nodes more explicitly. I'm not really sure what it's used for so I don't know the appropriate definition.

benjamn:
The Recast printer mostly matches on exact string `.type`s, so I agree it would be good for it to have a `Noop` printer case.

Just curious: how is `Noop` different from `EmptyStatement`? Could they be aliases for one another?

eventualbuddha:
I don't know how `Noop` differs from `EmptyStatement`. Maybe @kittens or @hzoo can comment?

hzoo:
I'm pretty sure we removed noop in v7 since we didn't really use it anywhere. ya https://github.com/babel/babel/pull/5970 - nvm it's just https://github.com/babel/babel/blob/593c1a08617eb3f14841fbabf07bda6d796065f6/packages/babel-types/src/definitions/misc.js#L4-L6, probably not different

eventualbuddha:
Hmm, not sure what the right course of action is. Should we just remove it here as well?

benjamn:
This library has other vestiges of no-longer-used AST types, and that's fine. They just sit there, mostly harmless, not causing any problems for the vast majority of ASTs that do not contain nodes of the bygone type. This principle is even more important when it comes to Babel- or Flow- or TypeScript-specific types: if you're not using a certain type, it shouldn't matter if `ast-types` has a definition for it. So I propose we wait to remove `Noop` until it becomes necessary to do so, though I am skeptical we will ever feel that strongly about it.

PR #249, opened 2018-02-05
Support TypeScript AST types, as parsed by Babylon 7

### LINKED ISSUE #213: Typescript support
Hello!
There is an ongoing effort to add TypeScript support to prettier. To do this, [typescript-eslint-parser](https://github.com/eslint/typescript-eslint-parser) is used to convert the AST from TypeScript into an estree compatible format. Due to some additional syntax, additional node types is required. It would be awesome to collect those here.

I have started defining some of them [here](https://github.com/Pajn/ast-types/commit/2e2a468e4283e69e1d11c71c25e2e61c4b4ee1cf). So I opened this issue to check with you first, if you are interested in integrating this at all and secondly, how to handle the development process.
To not create to big PRs and ease collaboration it would be great to integrate often which mean that these types will be extended and may see some churn until things stabilize. Should I hold off sending these to you now and instead include them in the prettier PR so that they can move from there at a later point in time or do you prefer to keep the development here?

Phoenixmatrix:
@benjamn 

This would resolve #213 

This isn't complete either, as I didn't have to understand all of the concepts to get the Recast printer to work. Some stuff like the builders don't have any arguments for the most part as I wasn't quite sure how they needed to be used in practice. The bases are also pretty much guts feeling/random as I was unfamiliar with the proper hierarchy, but it's a start to iterate on.

Prereq for benjamn/Recast#461


Phoenixmatrix:
Looks like I Have to take a closer look at the tests here (I've only been running the recasts one). Looks related to babylon parsing.

Phoenixmatrix:
ok, dee4c51 is when I broke the tests. WIll take a closer look

Phoenixmatrix:
Ok, all better. Builds are passing now.

benjamn:
@Phoenixmatrix Thanks so much for working on this!

I've taken the liberty of rewriting each commit to use two spaces per tab (my current preference, sorry that wasn't clear at all), preserving both the sequence of commits and your authorship/blame, and I squashed the optional `accessibility` fix into your original `TS: classes` commit. Of course, this required force-pushing to your branch, so you will probably want to rebase if you plan to push more commits. Sorry for any trouble that causes you.

My next plan of action is to go through [`babel/packages/babel-types/src/definitions/typescript.js`](https://github.com/babel/babel/blob/master/packages/babel-types/src/definitions/typescript.js) to get a sense for what's left, and make sure the types you've added so far seem consistent with what Babel/Babylon produce. I'll push more commits to this branch when/if I decide to make any additions.

Phoenixmatrix:
No problem about the indentation. I just tried to copy the current style as much as I could, but 2 spaces is my preference too, so that's even better.

I didn't know about the typescript babel-types definition! (I guess I should have considering it was brought up in the TS issue). That would have made things a lot easier than fighting with ast explorers. Learn something new every day!

Feel free to ask me to change anything if some major tweaks are required: I don't have any ego when it comes to my code. I won't be offended :)  Thanks for the help!

Phoenixmatrix:
Ahh yeah... looking at babel-types makes it a lot easier to figure out what's an expression, statement, and all the possible attributes. Whoops. I'm pretty sure I'm quite a bit off in some cases considering I guessed most of it by looking at the AST from sample code (I made most things bases as just Node since I didn't know any better)

benjamn:
Ok, I just finished a pass where I compared all the types in [`def/typescript.js`](https://github.com/Phoenixmatrix/ast-types/blob/ecce727197ca06a58c9e3eefe4dacbad1f33b555/def/typescript.js) to the corresponding types in [`babel/packages/babel-types/src/definitions/typescript.js`](https://github.com/babel/babel/blob/master/packages/babel-types/src/definitions/typescript.js). The next step will be checking for anything defined by `babel-types` that we haven't defined yet.

Ultimately, I would love to find some TypeScript libraries that exercise as much of the language as possible, so that we can do some deep whole-AST validation using the types we've defined (like we currently do [here](https://github.com/benjamn/ast-types/blob/69ffc00c091b4bd9056ac182a90786991e773e60/test/run.js#L219-L229)). At that point, I would be reasonably confident we've fully described everything that Babylon might generate.

Phoenixmatrix:
What about TypeScript itself (i think it's how the Prettier folks might have tested their stuff originally? Unless I'm remembering one). RxJS 5.* would be another good one. We might also want to try it on some of the React type definitions of definitelytyped, since type definitions use different constructs than regular code.

Of note is how in TS, jsx files (tsx) and regular ts files are kind of mutually exclusive (JSX won't work in normal ts files, and type assertions using the `<cast>foo` syntax won't work in tsx files). I don't think it matters much in the babylon implementation since it's handled via the jsx plugin overriding the conflicting tsx syntax, but might still be worth testing on some React projects.

benjamn:
Ok! The [`def/typescript.js`](https://github.com/Phoenixmatrix/ast-types/blob/87bcc0a3ac5e0b92b11a789c5e8722f43430bd37/def/typescript.js) definitions are now theoretically complete, or at least consistent with what's in [`babel/packages/babel-types/src/definitions/typescript.js`](https://github.com/babel/babel/blob/master/packages/babel-types/src/definitions/typescript.js).

I think I'm done working on this for today, so please feel free to look into the deep AST validation idea.

elliottsj:
Thanks for working on this @Phoenixmatrix + @benjamn; looking forward to this landing 😄 

Let me know if there's anything I can help with. If no one has started yet, I can starting looking into the deep-AST validation idea.

blaster151:
I'm really excited about this!  Thanks, @Phoenixmatrix and @benjamn!

benjamn:
Thanks to 7881d58283776308e3bb1aac04c0cd604b60a18c and https://github.com/benjamn/ast-types/pull/249/commits/36f1b26f2e854c675d13b644bf36bf40cb4c6229#diff-979039dad43a3c8533f0fc2663ebe659, we are now performing whole-AST validation on [all the same TypeScript test files](https://travis-ci.org/benjamn/ast-types/jobs/342862243#L695) that Babylon tests against. As always with this kind of validation, a number of small bugs were exposed (even bugs unrelated to TypeScript!), but I think I've fixed all of them.

Phoenixmatrix:
Awesome! Sorry you ended up needing to do most of the work in the end :)

Phoenixmatrix:
Woohoo!

PR #252, opened 2018-02-15
Allow building types with `.from` using an object

elliottsj:
With `.from`, you can be explicit about which fields you want to
populate, without having to deal with the positional arguments
of the normal builder.

See also:
https://github.com/facebook/jscodeshift/issues/180#issuecomment-365763352

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -533,77 +533,115 @@ module.exports = function () {
         // Override Dp.buildable for this Def instance.
         Object.defineProperty(self, "buildable", {value: true});
 
+        function addParam(built, param, arg, isArgAvailable) {
+            if (hasOwn.call(built, param))
+                return;
+
+            var all = self.allFields;
+            if (!hasOwn.call(all, param)) {
+                throw new Error("" + param);
+            }
+
+            var field = all[param];
+            var type = field.type;
+            var value;
+
+            if (isArgAvailable) {
+                value = arg;
+            } else if (field.defaultFn) {
+                // Expose the partially-built object to the default
+                // function as its `this` object.
+                value = field.defaultFn.call(built);
+            } else {
+                var message = "no value or default function given for field " +
+                  JSON.stringify(param) + " of " + self.typeName + "(" +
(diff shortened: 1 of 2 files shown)


elliottsj:
Here's my first stab at implementing the `.from` builder; let me know what you think

benjamn:
Looking forward to reviewing this weekend!

lib/types.js
```diff
@@ -533,77 +533,119 @@ module.exports = function () {
...
             }
+
+            return built;
+        }
+
+        // Calling .from on the builder function will construct an instance of the Def,
+        // using field values from the passed object. For fields missing from the passed object,
+        // their default value will be used.
+        builder.from = function (obj) {
+            var built = Object.create(nodePrototype);
```
benjamn: This object creation can come after the `self.finalized` check, though I guess the same could be said for the existing code in the `builder` function.

lib/types.js
```diff
@@ -533,77 +533,119 @@ module.exports = function () {
...
+                if (hasOwn.call(obj, param)) {
+                    addParam(built, param, obj[param], true);
+                } else {
+                    addParam(built, param, null, false);
+                } 
+            });
+
+            Object.keys(self.allFields).forEach(function (param) {
+                // Use the default value.
+                addParam(built, param, null, false);
```
benjamn: I think this loop also needs to check whether `obj[param]` is defined, so that `obj` can provide values for more than just the `.build` parameters, e.g.
```js
builders.identifier.from({ name: "foo", typeAnnotation: ... })
```
The `typeAnnotation` parameter isn't one of the `.build` parameters, but it seems like it should get checked and included, if provided.

lib/types.js
elliottsj: Fixed

lib/types.js
elliottsj: Ah, you're right. I think we don't even need to loop over `self.buildParams` either since any field in `self.allFields` could be specified in `obj`. Now updated.

PR #262, opened 2018-04-22
Add missing builder for TSTypeReference

theRobinator:
Looks like this was missed.


theRobinator:
Seems that the build is failing in the same way that all the other PRs' builds are, including the greenkeeper builds. I guess that's a bigger issue?

benjamn:
@theRobinator Right you are! I've fixed the test failures on `master` (see https://github.com/babel/babel/pull/7967 for a more permanent fix), and I think this PR is safe to squash-merge.

PR #263, opened 2018-04-23
Change field for DeclareModuleExports to TypeAnnotation

### LINKED ISSUE #261: Flow: DeclareModuleExports should take a TypeAnnotation
Hi! Firstly, I'm having a great time using `recast` and `ast-types` to code generate some flow library definitions, so thanks for all the dope work :)

I was attempting to build a module declaration that exports a CommonJS module a la [the flow docs](https://flow.org/en/docs/libdefs/creation/#toc-declaring-a-commonjs-module) when I ran into an issue.

```JS
declare module "some-module" {
  declare module.exports: {|
    someExport: number,
  |};
}
```

It looks like the definition that was added in [benjamn/ast-types/pull/162](https://github.com/benjamn/ast-types/pull/162/files#diff-b3e3de2d6bd0aa4938c8000843f7ed05R291) expects a `typeAnnotation` field that is a `def("Type")`.

From what I can see from the docs and the ast generated at flow.org/try, it should take a `typeAnnotation`, and a similar PR that tried to introduce `DeclareModuleExports` actually had this correct before being superseded: [benjamn/ast-types/pull/169](https://github.com/benjamn/ast-types/pull/169/files#diff-b3e3de2d6bd0aa4938c8000843f7ed05R294).

This is my first foray into `recast` and `ast-types` so please let me know if I'm misunderstanding! Happy to make a PR and test case if we're agreed this should be the expected behaviour!

ben-chin:
Closes #261 and adds a regression test case

### DIFF
--- a/def/flow.js
+++ b/def/flow.js
@@ -347,7 +347,7 @@ module.exports = function (fork) {
   def("DeclareModuleExports")
     .bases("Statement")
     .build("typeAnnotation")
-    .field("typeAnnotation", def("Type"));
+    .field("typeAnnotation", def("TypeAnnotation"));
 
   def("DeclareExportDeclaration")
     .bases("Declaration")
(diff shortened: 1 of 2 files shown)


benjamn:
Thanks @ben-chin!

PR #269, opened 2018-05-06
fix #267

### LINKED ISSUE #267: `npm test` fails to run on Windows (`test/run.sh` is not executed)
This is due to the particulars on Windows where the usual install of git-on-windows has a bash shell, but this is not the default command shell (for obvious reasons as then quite a lot of other unrelated things would fail), so any `npm` *script task* which executes a UNIX shell script should do this by invoking `bash -c <script_path>` instead of just invoking `<script_path>`.

GerHobbelt:
The adjusted npm script task `npm test` works on both Windows and UNIX boxes now. See #267.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   "license": "MIT",
   "main": "main.js",
   "scripts": {
-    "test": "test/run.sh"
+    "test": "bash -c test/run.sh"
   },
   "dependencies": {},
   "devDependencies": {


benjamn:
Why not just `bash test/run.sh` without the `-c`?

benjamn:
Also, I think I've fixed the cause of the test failures on `master`, so please feel free to rebase this PR.

GerHobbelt:
> Why not just `bash test/run.sh` without the `-c`?

`-c` is a hold-over from my MSDOS/4DOS stints back in the Days Of Antiquity; it works with `bash` as well and some times it flows out of my hands before my brain kicks in. Later, when reading it, it's a direct signal to me the shell is kickstarted for the sole purpose of executing the script and no more and don't get the urge to clean it up.

Hence: `-c` just info-leaks my dinosaur genes. Old habits, new hairline and all that jazz.  😈

PR #272, opened 2018-05-17
Add `?.` and `??` stage 1 nodes

qfox:
Add support for stage 1 proposals "Optional Chaining" (`?.`) and "Nullish Coalescing" (`??`) operators.

This adds the following nodes;

- `OptionalMemberExpression` (for `a?.b` and `a?.[b]`)
- `OptionalCallExpression` (for `a.b()` and `a?.[b]()`)

It adds an option to `LogicalExpression` to allow `??` as the `operator`.

Tests have been added.

### DIFF
--- a/def/es-proposals.js
+++ b/def/es-proposals.js
@@ -1,0 +1,33 @@
+module.exports = function (fork) {
+  fork.use(require('./core'));
+
+  var types = fork.use(require("../lib/types"));
+  var Type = types.Type;
+  var def = types.Type.def;
+  var or = Type.or;
+
+  var shared = fork.use(require("../lib/shared"));
+  var defaults = shared.defaults;
+
+
+  // https://github.com/tc39/proposal-optional-chaining
+  // `a?.b` as per https://github.com/estree/estree/issues/146
+  def("OptionalMemberExpression")
+    .bases("MemberExpression")
+    .build("object", "property", "computed", "optional")
+    .field("optional", Boolean, defaults["true"])
+
+  // a?.b()
+  def("OptionalCallExpression")
+    .bases("CallExpression")
+    .build("callee", "arguments", "optional")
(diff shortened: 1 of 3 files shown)


benjamn:
@qfox Thanks for this! I think I've fixed the cause of the test failures on `master`, so could you please rebase this PR?

Personal note: I'll be out of the country until Tuesday (May 22nd), so I probably won't be able to respond again until after that.

def/es-proposals.js
```diff
@@ -0,0 +1,33 @@
...
+
+  // https://github.com/tc39/proposal-optional-chaining
+  // `a?.b` as per https://github.com/estree/estree/issues/146
+  def("OptionalMemberExpression")
+    .bases("MemberExpression")
+    .build("object", "property", "computed", "optional")
+    .field("optional", Boolean, defaults["true"])
+
+  // a?.b()
+  def("OptionalCallExpression")
```
benjamn: How is this different from a `CallExpression` whose `callee` is an `OptionalMemberExpression`? Or can you do something like `f?()`?

qfox: No, this is more like `a?.b()`. Since the operator is short circuiting the AST needs to know whether the method call is optional as well.

qfox:
Sure, I'll rebase over the weekend.

def/es-proposals.js
qfox: Actually, I double checked this and I stand corrected. Yes, it's `a?.()`.
See https://github.com/tc39/proposal-optional-chaining#base-case

qfox:
My fork has been rebased and I think that automatically updates this PR? :)

qfox:
@benjamn Hey Ben. Any chance you could take a look at this? It's blocking updates to jscodeshift and it'd be great if it could be added. Let me know if you have other comments. Thanks!

PR #290, opened 2018-10-24
Correct capitalization of supertype and impltype

jbrown215:
The flow AST calls these fields `supertype` and `impltype` (as opposed to `superType` and `implType`). See the [ast tab](https://flow.org/try/#0PYBwhgjgrgpgBAFwJ4ngFQFxwN4F84C8OuA3EA) here.

This prevents tools like recast from traversing these fields, which means tools like jscodeshift won't codemod them appropriately. This was quite the bug to track down :D

### DIFF
--- a/def/flow.ts
+++ b/def/flow.ts
@@ -294,8 +294,8 @@ export default function (fork: Fork) {
     .build("id", "typeParameters", "impltype", "supertype")
     .field("id", def("Identifier"))
     .field("typeParameters", or(def("TypeParameterDeclaration"), null))
-    .field("implType", def("FlowType"))
-    .field("superType", def("FlowType"));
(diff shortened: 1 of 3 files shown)


jbrown215:
https://github.com/facebook/flow/blob/master/packages/flow-parser/test/custom_ast_types.js#L52-L58 for more confidence

benjamn:
Meanwhile almost every other field name is camel-cased! Sorry you had to hunt this down.

benjamn:
These changes are now published in `ast-types@0.12.1`, which is also available through `recast@0.17.1`. Note that both versions were published to npm using the `next` tag, rather than `latest`, though we will switch to `latest` soon.

Thanks for your patience @jbrown215!

PR #292, opened 2018-10-25
Add inexact flag to flow ObjectTypeAnnotation

jbrown215:
See https://github.com/facebook/flow/blob/master/packages/flow-parser/test/custom_ast_types.js#L181

### DIFF
--- a/gen/nodes.ts
+++ b/gen/nodes.ts
@@ -845,6 +845,7 @@ export interface ObjectTypeAnnotation extends Omit<FlowType, "type"> {
   properties: (K.ObjectTypePropertyKind | K.ObjectTypeSpreadPropertyKind)[];
   indexers: K.ObjectTypeIndexerKind[];
   callProperties: K.ObjectTypeCallPropertyKind[];
+  inexact: boolean | undefined;
   exact: boolean;
   internalSlots: K.ObjectTypeInternalSlotKind[];
 }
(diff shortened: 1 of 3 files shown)


benjamn:
Any reason why `inexact`-ness couldn't be expressed by setting the `exact` field explicitly to `false`?

jbrown215:
Yes-- the plan is to eventually get rid of the `exact` flag entirely in favor of `inexact`. The issue is that many tools today expect to be able to operate on the `exact` flag, so we can't omit it in the case of implicit inexactness. Since we can't omit it, `exact = false` is ambiguous, since it can either mean `{}` or `{...}`. When we get rid of the `{||}` syntax, there won't be a need for the exact flag anymore and we will get rid of it.

gen/nodes.ts
```diff
@@ -845,6 +845,7 @@ export interface ObjectTypeAnnotation extends Omit<FlowType, "type"> {
   properties: (K.ObjectTypePropertyKind | K.ObjectTypeSpreadPropertyKind)[];
   indexers: K.ObjectTypeIndexerKind[];
   callProperties: K.ObjectTypeCallPropertyKind[];
+  inexact: boolean | undefined;
```
benjamn: Now that we're generating TypeScript interfaces for all the AST definitions, you can see the consequences of changes immediately!

benjamn:
These changes are now published in `ast-types@0.12.1`, which is also available through `recast@0.17.1`. Note that both versions were published to npm using the `next` tag, rather than `latest`, though we will switch to `latest` soon.

Thanks for your patience @jbrown215!

PR #295, opened 2018-11-14
fix(scope): add TypeParameterDeclarations to scope.types

### LINKED ISSUE #294: TypeParameterDeclarations are not added to scope.types
For instance in
```js
async function withStatus<T>(fn: () => Promise<T>): Promise<T> {
```
`T` doesn't get added to the function's `scope.types`.

### Uhoh...
@benjamn In ECMAScript, only functions and catch clauses introduce new scopes.  But flow types throw a wrench into the system: the scope of a type parameter is limited to the declaring class, function, interface, or type alias.  This means `ast-types` is **currently unable to represent type parameter scope correctly**.

I propose that we solve this by attaching scopes to the paths for class, function, etc. nodes, but flag those scopes such that they only allow type parameters to be declared on them. 

### Context
I'm trying to use `ast-types` to locate all undeclared identifiers in JS/Flow code.  Right now it's not working on code that uses type parameters.

jedwards1211:
fix #294

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -180,6 +180,12 @@ module.exports = function (fork) {
         } else if (namedTypes.TypeAlias && namedTypes.TypeAlias.check(node)) {
             addTypePattern(path.get("id"), scopeTypes);
 
+        } else if (namedTypes.TypeParameterDeclaration && namedTypes.TypeParameterDeclaration.check(node)) {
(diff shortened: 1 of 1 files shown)


jedwards1211:
nvm, this doesn't work.  Need to tweak more things

PR #297, opened 2018-11-14
fix(scope): add TypeParameterDeclarations and InterfaceDeclarations to scope.types

### LINKED ISSUE #294: TypeParameterDeclarations are not added to scope.types

### LINKED ISSUE #296: Add Flow InterfaceDeclarations to scope.types
I can make a PR for this soon

jedwards1211:
fix #294
fix #296

### Important Note
With this change, `NodePaths` for `ClassDeclaration`s, `ClassExpression`s, `InterfaceDeclaration`s, and `TypeAlias`es will create their own `Scope`.  *But*, when `scanScope` sees that the node is one of those, rather than a `Program`, `Function`, or `CatchClause`, it will only scan and add `TypeParameter`s to the `scope.types` -- it won't touch `scope.bindings`.

### DIFF
--- a/lib/path.ts
+++ b/lib/path.ts
@@ -11,9 +11,9 @@ export interface Path<V = any> {
   __childCache: object | null;
   getValueProperty(name: any): any;
   get(...names: any[]): any;
-  each(callback: any, context: any): any;
-  map(callback: any, context: any): any;
-  filter(callback: any, context: any): any;
+  each(callback: any, context?: any): any;
+  map(callback: any, context?: any): any;
+  filter(callback: any, context?: any): any;
   shift(): any;
   unshift(...args: any[]): any;
   push(...args: any[]): any;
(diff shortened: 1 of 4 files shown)


jedwards1211:
@benjamn I just rebased this, can you have a look?

jedwards1211:
@benjamn okay I made this PR also handle TS types

benjamn (approved):
@jedwards1211 Thanks for your patience here, and for taking the initiative to add tests and support for TypeScript!

Issue #298, opened 2018-11-18
Cannot build identifier with TypeScript typeAnnotation


brieb:
Trying to build an identifier with a TypeScript typeAnnotation throws an error because the expected type of the type annotation is `TypeAnnotation | null`, as defined in the flow defs https://github.com/benjamn/ast-types/blob/f4a7c6db8e0eed1f25971d4fcbb84843c83ad615/def/flow.js#L209-L210

For example, if I parse `const x: string` and look at the AST for the identifier I see

```
{
  "type": "Identifier",
  "name": "x",
  "typeAnnotation": {
    "type": "TSTypeAnnotation",
    "typeAnnotation": {
      "type": "TSStringKeyword"
    }
  }
}
```

But if I try to create this with a builder like so

```
astTypes.builders.identifier.from({ name: 'x', typeAnnotation: astTypes.builders.tsTypeAnnotation(astTypes.builders.tsStringKeyword()) })
```

I get the error

```
Uncaught Error: {typeAnnotation: [object Object], loc: null, type: TSTypeAnnotation, comments: null} does not match field "typeAnnotation": TypeAnnotation | null of type Identifier
```

I'll work on a fix for this and submit a PR

benjamn:
Ahh, I think `def/typescript.js` needs to redefine the fields of the `Identifier` type, like `def/flow.js` is doing, to include a `typeAnnotation` field with type `TSTypeAnnotation | null`.

One wrinkle: we import both `def/flow.js` and `def/typescript.js` by default [in `main.js`](https://github.com/benjamn/ast-types/blob/master/main.js), so there's some risk here of one clobbering the other. I guess what we really want is a `typeAnnotation` field with type `TypeAnnotation | TSTypeAnnotation | null`, which may require adding some functionality to the type system (in `lib/types.js`) for getting the current type of a field, so the redefinition code can just add another `|` option to the existing type (if any), rather than blindly redefining the field.

Another option would be to introduce a super-type of `TypeAnnotation` and `TSTypeAnnotation`, but it feels weird to be merging the Flow and TypeScript type systems together like that.

Kind of annoying how Flow claimed a bunch of un-prefixed type names (like `TypeAnnotation`) whereas (the Babel implementation of) TypeScript makes it clear that `TSTypeAnnotation` is TypeScript-specific.

brieb:
> One wrinkle: we import both def/flow.js and def/typescript.js by default in main.js, so there's some risk here of one clobbering the other.

[...]

PR #299, opened 2018-11-18
Support building types with either Flow or TS type annotations

brieb:
Fixes issue: https://github.com/benjamn/ast-types/issues/298


brieb:
Looks like the [node 4 and 5 build failures](https://travis-ci.org/benjamn/ast-types/builds/456721909) are unrelated to this change -- see them happening [here](https://travis-ci.org/benjamn/ast-types/builds/452222065) as well.

PR #300, opened 2018-11-19
Convert repo to Typescript; Auto-generate types for defs

brieb:
Following up from our twitter thread: https://twitter.com/benjamn/status/1063979819149475841

Started down the path of making a d.ts files, but in combing through the code to make sure the types were correct, I increasingly felt like I'd have more confidence if we just converted the code to TS. Sounded like you were open to this idea from your twitter comment.

Goals:
- Convert all files from JS to TS
- Emit TypeScript declarations
- Auto-generate types for types/builders/visitor, as exported from main
- Preserve all runtime behavior / only modify the code in the type space
- Backwards compatibility
  - Keep folder structure for emitted JS, to not change import paths for users
  - Keep `module.exports =` by using TypeScript's `export =`, so default exports don't change

The results are pretty awesome!

image


brieb:
The types are pretty loose, but are at least correct. We'll want to narrow these types over time.

def/e4x.ts
```diff
@@ -83,4 +87,4 @@ module.exports = function (fork) {
         .bases("XML")
         .field("target", String)
         .field("contents", or(String, null));
```
brieb: This file is indented with 4 spaces, whereas most are indented with 2. Have you thought about enabling prettier for this repo?

benjamn: I've been slowly shifting to 2 spaces per tab as I make changes to files, and I haven't touched this one in a while. Now seems like a fine time to reindent this code!

You're welcome to use any formatting tool you like when working on a PR, but I'd rather not add Prettier to the workflow for this repository. The readability of `def(...).bases(...).field(...)` expressions often benefits from idiomatic line breaks and indentation, I've found, and auto-formatting tools would only interfere.

.gitignore
```diff
@@ -1,3 +1,12 @@
...
 test/data/typescript-compiler
+
+# Ignore TypeScript-emitted files
+*.js
+*.js.map
+*.d.ts
+*.d.ts.map
+# Except...
+!/types/**/*.d.ts
+!babel-register.js
```
brieb: Didn't see a good way around this, given that we wouldn't want to change the import paths. Open to ideas

benjamn:
Since it seems like you're still pushing commits, I will wait for your signal before I start reviewing. Just let me know if anything is blocking you. I have a feeling I'm going to learn a lot about TypeScript by the end of this!

brieb:
Sorry, wasn't sure when you were going to take a look and noticed some things when I was integrating the types with recast. And had an idea on how to do the build, such that we didn't need to use `export =`, which made it hard to export types. Evolved a bit more than I originally expected when I put up the PR originally. 🙂 

Going to try writing jscodeshift types from the types emitted from ast-types and recast and make sure everything works out. So, only thing I'm planning on maybe touching are some type exports. I'll let you know when that settles.

I feel like I've learned so much about how all this works.

brieb:
YEEEESSSS!!!!

jscodeshift-ts

Done pushing now - PTAL. 👍

.gitignore
benjamn: I adopted your strategy from https://github.com/benjamn/recast/pull/554#issuecomment-440975441 of just using `tsc` and https://www.npmjs.com/package/ts-add-module-exports, so at least the `!babel-register.js` part of this is resolved (since we're not using Babel anymore).

benjamn:
Tests are passing now!

Some notes:
- Thanks for leaving your branch open for me to push some additional commits.
- I've merged `master` (which contains #299) into this PR. There were some merge conflicts, naturally, but I think I resolved them all correctly.
- No longer running tests in Node 4 and 5, per previous [discussion](https://github.com/benjamn/recast/pull/554#issuecomment-440470682).
- I liked [your idea](https://github.com/benjamn/recast/pull/554#issuecomment-440975441) of using `tsc` and http://npmjs.org/package/ts-add-module-exports instead of Babel, so I decided to try that with this PR. I think it worked!
- I still want to get to the bottom of the problem I outlined in this commit: https://github.com/benjamn/ast-types/pull/300/commits/c6e31a4edb131df886156bb4ae1084b07fb8d665
- That [gif](https://github.com/benjamn/ast-types/pull/300#issuecomment-440499675) is really awesome!

def/babel-core.ts
```diff
@@ -2,6 +2,7 @@ import { Fork } from "../types";
 import typesPlugin from "../lib/types";
 import sharedPlugin from "../lib/shared";
 import es7Def from "./es7";
+import * as N from "../gen/nodes";
```
brieb: > Do you think it's okay to refer to generated code in these def/*.ts files? Is there a better way to do this?

Was trying to avoid consuming any of the gen exports outside of `/gen` and `main.ts` because the type is dependent on which defs are loaded. Like for `NumericLiteral` below, suppose some other def adds some fields on it. Since gen is based on the defs used in main, if someone uses fork and doesn't provide the same defs, the type will be wrong.

Bummer.

Some Options:
1) Say the case I explained above would be rare and just use gen

2) Run the gen on a def-by-def basis and import that type here. And, separately keep running the overall gen to be exported by main.

3) Define an interface here to be used in `getDefault` and extended by the gen node type somehow. Though it would be weird for some types to have interfaces and others not. And annoying to keep the interface and def in sync.

4) Use `any`. No type safety. :/

5) Use an inline type for the fields used in `getDefault`. No real type safety since we could say the type is anything - there's nothing checking that the inline interface actually matches the fields.

---

Option 2 doesn't seem so bad. And makes it so the only fields available are the fields defined in the same def. So we don't get any accidental dependencies added by other defs.

package.json
```diff
@@ -32,7 +32,7 @@
     "tsc": "tsc --noEmit",
     "mocha": "test/run.sh",
     "test": "npm run gen && npm run tsc && npm run mocha",
-    "clean": "script/run.sh clean",
+    "clean": "ts-emit-clean",
```
brieb: 👍

package.json
```diff
@@ -28,14 +28,12 @@
...
-    "gen": "script/run.sh gen-types",
+    "gen": "ts-node --transpile-only script/gen-types.ts",
     "tsc": "tsc --noEmit",
     "mocha": "test/run.sh",
     "test": "npm run gen && npm run tsc && npm run mocha",
     "clean": "ts-emit-clean",
-    "build": "npm run build:dts && npm run build:js",
-    "build:dts": "tsc --emitDeclarationOnly",
-    "build:js": "babel . --out-dir \".\" --extensions \".ts\" --ignore \"**/*.d.ts\",\"node_modules\",\"test/data\"",
+    "build": "tsc && ts-add-module-exports",
```
brieb: 👍

def/type-annotations.ts
```diff
@@ -52,9 +52,9 @@ export default function (fork: Fork) {
                 def("TSTypeParameterInstantiation"),
                 null),
              defaults["null"])
-      .field("implements",
-             or([def("ClassImplements")],
-                [def("TSExpressionWithTypeArguments")]),
-             defaults.emptyArray);
+      // .field("implements",
```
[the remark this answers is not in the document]
brieb: Turns out it works if the class is immediately returned.

```
// ok
export function foo() {
  return {
    foo: class Foo {
      someMethod() {
        return false;
      }
    },
  }
}

// not ok - Return type of exported function has or is using private name 'Foo'. [4060]
export function foo() {
  class Foo {
    someMethod() {
      return false;
    }
  }

  return { foo: Foo };
}
```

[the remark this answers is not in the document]
benjamn: Interesting!

brieb:
Updated the PR to use Type subtypes (OrType, ArrayType, ObjectType) for codegen instead of relying on the string encoding.

Started converting things to classes but it got pretty messy. And the constraint of having to return the class immediately didn't work in some cases where the class needed to be referenced before the return statement. So, seemed cleanest for now to leave things as is. Perhaps we could better clean this up using an alternative plugin structure. Not sure yet what that would look like exactly though.

As far as [generating types per def](https://github.com/benjamn/ast-types/pull/300#discussion_r236383932), that would complicate the generation code and multiply the number of codegen-ed files by N * (1 for nodes.d.ts + 1 for kinds.d.ts + ...). Not sure if that added complexity is warranted for the minimal number of bugs it would help prevent. Thoughts?

benjamn:
I don't have any good ideas yet for generating specific types per def, but I'm mostly OK with the possibility of having too many fields on a type, or too many options in a union, as long as the types are correct for the default setup.

I think I have a solution that allows `Type` to be an actual class, without triggering the error about private types. I was hoping to push that work tonight, but I still need to get a few tests to pass, and I've run out of time this evening. In the meantime, just be aware that changes are coming (especially in `lib/types.ts`), in case you wanted to push any other changes, though I'm happy to rebase if necessary.

lib/types.ts
```diff
@@ -256,151 +328,13 @@ export default function typesPlugin(_fork?: Fork) {
   var isNull = defBuiltInType<null>(null, "null");
   var isUndefined = defBuiltInType<undefined>(void 0, "undefined");
 
-  // There are a number of idiomatic ways of expressing types, so this
-  // function serves to coerce them all to actual Type objects. Note that
-  // providing the name argument is not necessary in most cases.
-  function toType(from: any, name?: any): AnyType {
```
benjamn: This `toType` helper has become `Type.from`.

script/gen-types.ts
```diff
@@ -280,10 +216,7 @@ const out = [
                               return field1.name < field2.name ? -1 : 1;
                             })
                             .map(fieldName => {
-                              const field = typeDef.allFields[fieldName];
-
-
-                              return getPropertySignature(field);
+                              return typeDef.allFields[fieldName].getPropertySignature(b);
```
benjamn: Since `Field` no longer has access to the `builders` object, I needed to pass the current `b` object into `getPropertySignature` here.

benjamn:
@brieb Ok, I'm done pushing what I was working on yesterday. I hope the commit messages and in-line comments are helpful for explaining the strategy. Feedback and/or follow-up commits welcome!

lib/types.ts
```diff
@@ -230,6 +255,31 @@ export default function typesPlugin(_fork?: Fork) {
...
       return this.name;
     }
+
+    getTypeAnnotation() {
+      if (typeof this.name !== "string") {
+        return builders.tsAnyKeyword();
+      }
+
+      if (hasOwn.call(namedTypes, this.name)) {
+        // TODO Make this work even if TypeScript types not used?
```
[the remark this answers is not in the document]
brieb: Wonder if we'd want to call this `getTSTypeAnnotation` in case we do ever support Flow type generation.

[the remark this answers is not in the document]
brieb: (to keep things consistent with the name of the type declarations)

def/typescript.ts
```diff
@@ -15,7 +15,7 @@ export default function (fork: Fork) {
   var def = types.Type.def;
   var or = types.Type.or;
   var defaults = fork.use(sharedPlugin).defaults;
-  var StringLiteral = new types.Type(function (value, deep) {
```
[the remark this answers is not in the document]
brieb: Makes sense. Now that we're exporting something that's not constructable, I wonder if it would simplify things to use plain objects and interfaces vs abstract classes and concrete subclasses.

lib/types.ts
```diff
@@ -230,6 +255,31 @@ export default function typesPlugin(_fork?: Fork) {
...
+        return builders.tsTypeReference(builders.tsQualifiedName(
+          builders.identifier("K"), // TODO Don't hard-code this.
```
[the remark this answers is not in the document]
brieb: Hm. Feeling kind of wary about coupling the lib files to the gen files... I'll try experimenting with possible ways to decouple them more.

brieb:
Made some changes to decouple the core Type class from the TS annotations. The motivation was to avoid bloating the core Type class with this extra behavior since it's pretty custom. This would also apply if we wanted to add other behaviors to Types in the future unrelated to the core assert/check behaviors. Instead of having subclasses, each Type has an underlying private type that is a union of the possible subtypes. This makes the Type class more of a utility wrapper around these internal types. By extracting the underlying type representation in this way, I could pull out the type annotations logic into a separate plugin and without making subclasses for each subtype. And, from there it was easy to add a way to configure the type annotations, so we don't have to hard code `K.SomeKind`. Though it still feels a bit weird to have the type annotation logic split across gen-types and lib/type-annotations. Should lib be responsible for more type annotations and gen-types just consumes that and prints it to a file?

brieb:
@benjamn What do you think of merging this at one of these commits and iterate on the types in follow-up PRs?

- 5d4d7076c367be2a62e1afc2b0125b3e067ab267
- 2f466c5f0dbc85a72b47b4768f146a557144cee6
- 0a8ec271215c6bf3962414f6f390d27e8fb051a9

brieb:
I was able to repro the build failures locally by rm -rf ing test/data and nvm use 7
The issue is that defs for ClassPrivateMethod are missing
I'll add those to fix the build

lib/types.ts
```diff
@@ -147,13 +189,170 @@ export interface Builder {
   from(obj: { [param: string]: any }): ASTNode;
 }
 
+type TypeKind = TypeUnion<any>["kind"];
+type TypesByKind = { [K in TypeKind]: Extract<TypeUnion<any>, { kind: K }> };
+type FromForType<T> = Omit<T, "kind" | typeof __typeBrand>;
+type TypeBuilders = { [K in TypeKind]: (from: FromForType<TypesByKind[K]>) => TypesByKind[K] }
+const typeBuilders: TypeBuilders = {
+  ArrayType(from) {
+    return { kind: "ArrayType", ...from } as ArrayType<any>;
```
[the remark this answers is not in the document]
brieb: The type for `from` omits `kind` so it would be a compiler error to pass `kind` to `from`.

types.ts
```diff
@@ -7,3 +7,5 @@ export type Plugin<T> = (fork: Fork) => T;
 export type Def = Plugin<void>;
 
 export type Omit<T, K> = Pick<T, Exclude<keyof T, K>>;
+
+export type Constructor<T = {}> = new (...args: any[]) => T;
```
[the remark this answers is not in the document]
brieb: ah. i had indeed experimented with using the mixin pattern from that page. i must have forgotten to remove the type after experimenting. i'll remove it.

gen/kinds.ts
```diff
@@ -0,0 +1,283 @@
...
...N.TSTypeReference | N.TSAsExpression | N.TSNonNullExpression | N.TSAnyKeyword | N.TSBooleanKeyword | N.TSNeverKeyword | N.TSNullKeyword | N.TSNumberKeyword | N.TSObjectKeyword | N.TSStringKeyword | N.TSSymbolKeyword | N.TSUndefinedKeyword | N.TSUnknownKeyword | N.TSVoidKeyword | N.TSThisType | N.TSArrayType | N.TSLiteralType | N.TSUnionType | N.TSIntersectionType | N.TSConditionalType | N.TSInferType | N.TSTypeParameter | N.TSParenthesizedType | N.TSFunctionType | N.TSConstructorType | N.TSDeclareFunction | N.TSDeclareMethod | N.TSMappedType | N.TSTupleType | N.TSRestType | N.TSOptionalType | N.TSIndexedAccessType | N.TSTypeOperator | N.TSIndexSignature | N.TSPropertySignature | N.TSMethodSignature | N.TSTypePredicate | N.TSCallSignatureDeclaration | N.TSConstructSignatureDeclaration | N.TSEnumMember | N.TSTypeQuery | N.TSTypeLiteral | N.TSTypeAssertion | N.TSEnumDeclaration | N.TSTypeAliasDeclaration | N.TSModuleBlock | N.TSModuleDeclaration | N.TSImportEqualsDeclaration | N.TSExternalModuleReference | N.TSExportAssignment | N.TSNamespaceExportDeclaration | N.TSInterfaceBody | N.TSInterfaceDeclaration | N.TSParameterProperty | N.OptionalMemberExpression | N.OptionalCallExpression;
```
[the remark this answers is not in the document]
brieb: we use recast pretty print to print this file. should we improve the formatting for union types there?

brieb:
Welcome back! Hope you had an awesome trip. No worries about the delay. It has been a lot of fun working with you on this. 😃

gen/kinds.ts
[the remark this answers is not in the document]
benjamn: Oh that's right! Yeah, there's definitely some Recast logic to line-break other kinds of syntax that can get too long for one line, so we can easily address that downstream.

lib/type-annotations.ts
```diff
@@ -0,0 +1,106 @@
...
+export interface TSTypeAnnotatorConfig {
+  getReferenceToKind(typeName: string): any;
+}
+
+export default function typeAnnotationsPlugin(fork: Fork) {
+  const { builders, namedTypes } = fork.use(typesPlugin);
+
+  function createTSTypeAnnotator(config: TSTypeAnnotatorConfig) {
+    function getTSTypeAnnotation(type: Type<any>): any {
+      const privateType = type.__type;
```
brieb: How are you feeling about this approach? It did feel weird to expose this private field in this way. But, I liked how it allowed us to move the type annotation logic into a separate isolated plugin.

Alternatively, we could rename `PrivateType` to `Type` and use that as the common `type` type and expose functions that take in that type like assert/check/getTSTypeAnnotation etc. like `function check<T>(type: Type<T>, value: any, deep?: Deep): value is T { ... }`. Would make things more object-centric and functional vs classes.

brieb: Made a pass at unifying the Type class and tagged union Type approach here https://github.com/benjamn/ast-types/pull/300/commits/20a5ae2d0085a8966e5addbd30ba013053c938b3

How does that feel to you?

lib/type-annotations.ts
```diff
@@ -0,0 +1,106 @@
+import { Fork } from "../types";
+import typesPlugin, { Field, Type } from "./types";
+import { assertNever } from "./utils";
+
+const Op = Object.prototype;
+const hasOwn = Op.hasOwnProperty;
+
+export interface TSTypeAnnotatorConfig {
+  getReferenceToKind(typeName: string): any;
```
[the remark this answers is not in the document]
brieb: It was also unclear to me what boundary we were trying to draw bringing in only part of the type generation into lib and leaving other parts in gen-types.ts here https://github.com/benjamn/ast-types/pull/300/commits/c6daab881535a26918bb3aeb01f85e91c0684ca0

Can you explain a bit more why this code should be in lib vs it all being in gen-types?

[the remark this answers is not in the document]
benjamn: You're right, it was awkward for the `gen-types.ts` code that was generating the type declarations to have to call `field.type.getTypeAnnotation()`, blindly trusting some arbitrary code from another module. I think that division was just an artifact/accident of object-oriented style, since I felt like the individual type classes (`ArrayType`, `OrType`, etc.) should only be responsible for generating the type annotations that they represented. Moving all this logic into `gen-types.ts` makes sense.

benjamn:
If you think this is ready to merge, maybe we can get away with bumping the `ast-types` version and publishing the changes to npm using the `beta` or `next` tag, so that other packages (e.g. `recast`) can try this stuff out before it becomes the default version?

I don't *think* there are any major breaking changes, though this PR does seem like a pretty substantial rewrite. Maybe just a minor version bump?

lib/types.ts
```diff
@@ -1,74 +1,150 @@
...
 
-type Deep = boolean | ((type: Type<any>, value: any) => void);
+declare const __typeBrand: unique symbol;
 
-export interface Type<T> {
-  toString(): string;
-  check(value: any, deep?: Deep): value is T;
-  assert(value: any, deep?: Deep): value is T;
+abstract class BaseType<T> {
+  readonly [__typeBrand]: T;
```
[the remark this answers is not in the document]
brieb: Unused. Removed it here https://github.com/benjamn/ast-types/pull/300/commits/c64137d38ab4dccbd676dde04520fa7554c25425

gen/nodes.ts
```diff
@@ -0,0 +1,2310 @@
...
+
+export interface SourceLocation extends ASTNode {
+  type: "SourceLocation";
+  start: K.PositionKind;
+  end: K.PositionKind;
+  source?: string | null;
+}
+
+export interface Node extends ASTNode {
+  loc?: K.SourceLocationKind | null;
```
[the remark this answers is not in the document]
brieb: Changed it to extend base types here https://github.com/benjamn/ast-types/pull/300/commits/3aef38d5f9adf615afc1d49aec7b979954b47b21

benjamn:
Any last reservations about merging this PR in its current state?

brieb:
I have a couple more changes. Rebasing on top of your changes now.

brieb:
As far as versioning, looks like ast-types is versioned as 0.x.x. Feels like this would warrant v0.12.0 given the extensiveness of the changes. Also, `Type` is no longer a class and there were changes to the defs, which could be considered breaking changes.

script/gen-types.ts
```diff
@@ -233,11 +233,7 @@ const out = [
                         typeAnnotation: b.tsTypeLiteral.from({
                           members: Object.keys(typeDef.allFields)
                             .filter(fieldName => fieldName !== "type")
-                            .sort((fieldName1, fieldName2) => {
-                              const field1 = typeDef.allFields[fieldName1];
-                              const field2 = typeDef.allFields[fieldName2];
-                              return field1.name < field2.name ? -1 : 1;
-                            })
+                            .sort() // Sort field name strings lexicographically.
```
brieb: Huh. The previous code doesn't make any sense. Why use the `fieldName` to get the `field.name`... Wonder if that was left over from a refactor. Thanks for cleaning that up.

lib/types.ts
```diff
@@ -119,7 +119,7 @@ class OrType<T> extends BaseType<T> {
   }
 }
 
-class PredicateType<T> extends BaseType<T> {
+export class PredicateType<T> extends BaseType<T> {
```
brieb: Instead, we could change the return type of `defBuiltInType` to `Type<T>` instead of `PredicateType<T>` to hide the implementation detail that the built-in types are `PredicateType`s internally. I'll go ahead and make that change.

brieb:
Done pushing. I'll try updating the tar-ed build in the recast PR as a quick sanity check and report back.

brieb:
Update: Investigating some test failures when consuming this version of ast-types in recast.

benjamn:
If the failures have anything to do with `ClassPrivateName`, that's probably just because we need to implement a printer case for that AST type.

brieb:
Looks like we dropped the `arrayOf` `Type` method [here](https://github.com/benjamn/ast-types/pull/300/commits/d25435ce0090cc5a32f31a14bde4f06254f071ea). Added it back [here](https://github.com/benjamn/ast-types/pull/300/commits/dfce3b4ae86cf7be39689d6956f63024e782c8ea).

brieb:
Also, what resource to you use when creating defs? I had been looking at https://github.com/babel/babel/blob/master/packages/babel-parser/ast/spec.md

brieb:
Made some updates to the defs to while adding printer support for the new types [in the recast PR](https://github.com/benjamn/recast/pull/554).

One thing I noticed is that builder params don't invoke the defaultFn if `undefined` is provided as an explicit argument. They only take effect if they are omitted. This is problematic if a consumer wants to provide an argument that follows a param with a default. Is this expected behavior?

brieb:
Here's a diff of just the defs in case that's helpful.
https://gist.github.com/brieb/dd8d9d6d1b352ef079172c601eba3e06

benjamn:
> One thing I noticed is that builder params don't invoke the defaultFn if `undefined` is provided as an explicit argument. They only take effect if they are omitted. This is problematic if a consumer wants to provide an argument that follows a param with a default. Is this expected behavior?

I agree that treating undefined values the same as omitted arguments (and calling the `defaultFn`) is a good idea. Happy to take a PR to fix that, at least for builder functions with positional arguments. I'm not sure about the `.from({...})` helper, since it's easier to omit properties with that style. But perhaps the same rules should apply everywhere, for consistency.

benjamn:
Welcome back! Glad you were able to disconnect. While the winter holidays can be a good time for me to catch up on non-work open source projects like this one, this time I had a wedding to plan (my own, Dec 29), so that definitely didn't happen. The open source, I mean.

I'd love to get this PR merged and released so that any follow-up changes can happen in separate PRs. Just pushed a few commits to see if tests were passing, and to fix the ones that started failing due to recent AST innovations by the Babel team.

I had one question about your last comment, but I'll put that in the relevant code.

Cheers! 🌲🎁🎉

gen/nodes.ts
```diff
@@ -14,7 +14,7 @@ export interface SourceLocation {
 }
 
 export interface Node extends Printable {
-  type: string;
+  type: "Node";
```
benjamn: > Not all nodes have `type`, just buildable ones. So, the generated types aren't accurate there.

Do you mean that non-buildable types should have `type: string` like before, since (for example) `"Node"` is not a type that any parser would produce? Would that cause any problems for the type narrowing behavior?

brieb: Yes, exactly. The types felt misleading in that case / not true to the underlying runtime behavior. For example, we'd never have a node with the type "Node" at runtime.

Just added a commit https://github.com/benjamn/ast-types/pull/300/commits/2d12dcb938c72ee8b1863334e91bb250dfe3bcf3 that changed it back to using the type's "own" fields to make the generated types more accurate.

Then, followed that up with a commit https://github.com/benjamn/ast-types/pull/300/commits/02f7b86b276b1f813aaaa1913e60943d0eb681dd that adds an export for a union type of all buildable types. These are the only ones that actually would have been constructed at runtime.

The type narrowing works great at the top-level.
image

The issue is if any "kind" (as in gen/kinds) unions something that introduces `type: string` the narrowing breaks.
For example, `CatchClause` has `param: K.PatternKind | null;`.
`type PatternKind = N.Pattern | N.Expression | ...`
Where `export interface Pattern extends Node {}` and `export interface Node extends Printable { type: string; ... }`
image

Here is a self-contained demonstration of this behavior [on the TypeScript Playground](http://www.typescriptlang.org/play/index.html#src=type%20Omit%3CT%2C%20K%3E%20%3D%20Pick%3CT%2C%20Exclude%3Ckeyof%20T%2C%20K%3E%3E%3B%0D%0A%0D%0Ainterface%20A%20%7B%0D%0A%20%20%20%20type%3A%20string%3B%0D%0A%7D%0D%0A%0D%0Ainterface%20B%20extends%20A%20%7B%0D%0A%20%20%20%20type%3A%20%22B%22%3B%0D%0A%20%20%20%20b%3A%20string%3B%0D%0A%7D%0D%0A%0D%0Ainterface%20C%20extends%20A%20%7B%0D%0A%20%20%20%20type%3A%20%22C%22%3B%0D%0A%20%20%20%20c%3A%20string%3B%0D%0A%7D%0D%0A%0D%0Ainterface%20D%20extends%20A%2C%20Omit%3CB%2C%20%22type%22%3E%20%7B%0D%0A%20%20%20%20type%3A%20%22D%22%3B%0D%0A%20%20%20%20d%3A%20string%3B%0D%0A%7D%0D%0A%0D%0Atype%20U1%20%3D%20A%20%7C%20B%20%7C%20C%20%7C%20D%3B%0D%0Atype%20U2%20%3D%20B%20%7C%20C%20%7C%20D%3B%0D%0A%0D%0Afunction%20fn1(u%3A%20U1)%20%7B%0D%0A%20%20%20%20if%20(u.type%20%3D%3D%3D%20%22C%22)%20%7B%0D%0A%20%20%20%20%20%20%20%20%2F%2F%20type%20does%20not%20narrow%20because%20of%20%60type%3A%20string%60%0D%0A%20%20%20%20%7D%0D%0A%7D%0D%0A%0D%0Afunction%20fn2(u%3A%20U2)%20%7B%0D%0A%20%20%20%20if%20(u.type%20%3D%3D%3D%20%22B%22)%20%7B%0D%0A%20%20%20%20%20%20%20%20%2F%2F%20type%20narrows%20to%20%60B%60%0D%0A%20%20%20%20%20%20%20%20console.log(u.b)%3B%0D%0A%20%20%20%20%7D%20else%20if%20(u.type%20%3D%3D%3D%20%22C%22)%20%7B%0D%0A%20%20%20%20%20%20%20%20%2F%2F%20type%20narrows%20to%20%60C%60%0D%0A%20%20%20%20%20%20%20%20console.log(u.c)%3B%0D%0A%20%20%20%20%7D%20else%20if%20(u.type%20%3D%3D%3D%20%22D%22)%20%7B%0D%0A%20%20%20%20%20%20%20%20%2F%2F%20type%20narrows%20to%20%60D%60%0D%0A%20%20%20%20%20%20%20%20console.log(u.b%2C%20u.d)%3B%0D%0A%20%20%20%20%7D%0D%0A%7D%0D%0A).

I noticed that the `@babel/types` uses the union-ed node type's `type` field - that might have been added to address this situation. https://gist.github.com/brieb/e312c8f2565c4b2cde5996e1f648780e#file-index-d-ts-L41

So, if we change the `Node` interface to have `type: ASTNode["type"]` we can retain the string literal union / ability to narrow.
image

So, I added another commit to use `type: ASTNode["type"]` rather than `type: string` https://github.com/benjamn/ast-types/pull/300/commits/493d64fb0515e238a253de4bb2854247fa9a79ea

test/ecmascript.ts
```diff
@@ -264,20 +268,23 @@ describe("whole-program validation", function() {
   this.timeout(20000);
 
   // TODO(brieb): do we need these?
```
brieb: Looks like this is the only remaining `TODO(brieb)` in this PR. Before, we ran `validateECMAScript` on this repo's source files ("main.js", "lib/shared.js", etc). How should we address these test cases, now that the source code is .ts? Should we run it on the emitted files, add different more targeted test cases as new fixtures, remove them, something else?

gen/nodes.ts
brieb: I take that last part back about being able to narrow the nested `CatchClause` `param`. Using `type: ASTNode["type"]` doesn't actually work as I'd expected. Going to fix that up and push a commit.

```
function exampleFn(node: ASTNode) {
  if (node.type === "CatchClause") {
    if (node.param) {
      if (node.param.type === "BlockStatement") {
        // Oops. `node.param` is still `CatchClause.param: Expression | Pattern | XMLAnyName | XMLQualifiedIdentifier | XMLFunctionQualifiedIdentifier | XMLAttributeSelector | XMLFilterExpression | XMLElement | XMLList`
      }
    }
  }
}
```

test/ecmascript.ts
benjamn: ~I think it would be useful to run `namedTypes.Program.assert(ast, true)` on some TypeScript ASTs, to assert that everything in the AST matches the expected types. Basically the same as `shared.validateECMAScript`, except it would only use the Babel TypeScript parser, not Esprima.~

Actually, scratch that! It looks like we already do a deep assertion for every file we process in `tests/typescript.ts` [here](https://github.com/benjamn/ast-types/blob/493d64fb0515e238a253de4bb2854247fa9a79ea/test/typescript.ts#L145), which runs over the whole https://github.com/Microsoft/TypeScript/tree/master/src/compiler codebase.

In light of that, I'm pretty sure the commented-out tests are redundant, and we can just remove them.

brieb:
Cool - I think we're good to go as long as you think this change https://github.com/benjamn/ast-types/pull/300#discussion_r246196883 / https://github.com/benjamn/ast-types/pull/300/commits/2483cd2d56ce3e33697d5ce7a10d7a21fb694074 makes sense!

benjamn:
Makes sense to me!

So… if `ASTNode` is a union of all buildable node interface types, and each `*Kind` type is also a union of (some) buildable node interface types, then I guess `ASTNode` is equivalent to the union of all `*Kind` types? I think the way you're currently defining `ASTNode` is simpler/better, but it seems like it could also be written
```ts
export type ASTNode = K.PrintableKind | K.NodeKind | K.FileKind | ...
```
If I'm understanding correctly, it's kinda neat how nodes and kinds are related via `ASTNode`.

I noticed one small oddity: `SourceLocation` currently has `type: "SourceLocation"`, even though parsed `{ start, end }` objects generally don't have a `type` field. This could be misleading if you wrote code like
```ts
if (node.type === "SourceLocation") {
  doSomething(node.start, node.end);
}
```
since no `node` would ever match that case.

However, that's just because I made `SourceLocation` buildable [here](https://github.com/benjamn/ast-types/blob/1d1a0da183e913c0e0eb448fe3fd9b5a53e81cdb/def/core.ts#L32). Honestly I don't think we need a builder function for `SourceLocation` nodes, or `Position` nodes, for that matter. I'll push a commit to fix that.

brieb:
Merged! Woo!

Yes, it should be the case that the union of all kinds resolves to the same union as `ASTNode` with one exception - some of the `XML*` types didn't have buildable subtypes, so I just had it fall back to the supertype's node type.
https://github.com/brieb/ast-types/blob/fd95109b3e8e4b3e4719a957b931a5805a2983a4/script/gen-types.ts#L40-L49
https://github.com/brieb/ast-types/blob/fd95109b3e8e4b3e4719a957b931a5805a2983a4/gen/kinds.ts#L92-L110

Here is the full list of supertypes without buildable subtypes (cases where `buildableSubtypes.length === 0`)
```
XMLDefaultDeclaration
XMLAnyName
XMLQualifiedIdentifier
XMLFunctionQualifiedIdentifier
XMLAttributeSelector
XMLFilterExpression
XML
XMLElement
XMLList
XMLEscape
XMLText
XMLStartTag
XMLEndTag
XMLPointTag
XMLName
XMLAttribute
XMLCdata
XMLComment
XMLProcessingInstruction
```

benjamn:
Oh geez, we should probably stop loading `def/e4x.ts` by default. E4X was a version of JavaScript before ES5 that was abandoned by TC39. The big new feature (never realized) was XML literals, so the original Mozilla Parser API (which inspired the Esprima parser) still included those `XML*` types, which is the only reason I included them in this project.

benjamn:
Ok, I've published version 0.12.0 (with this PR included) to npm using the `next` tag. That should make `ast-types@0.12.0` easier to consume in Recast and elsewhere. Happy to promote it to `latest` whenever it seems ready.

brieb:
Awesome! Just updated https://github.com/benjamn/recast/pull/554 to use that version.

PR #309, opened 2019-01-27
Add definition for `TSImportType` (TypeScript 2.9+).

GerHobbelt:
The tests fail due to this type definition missing when you upgrade to TypeScript 2.9+ as that one produces an AST which includes TSImportType nodes (when using `import(...)`).

Run `test/run.sh` to update the test fixtures, update to tsc 2.9+ and you'll observe this problem too.


benjamn:
Thanks for tracking down the test failures and tackling this definition!

[Here](https://github.com/babel/babel/blob/738060ebfa7ac133d7dd6a2590835acaa08f15f3/packages/babel-parser/src/plugins/typescript.js#L223-L243)'s the `@babel/parser` method for this type:
```ts
tsParseImportType(): N.TsImportType {
  const node: N.TsImportType = this.startNode();
  this.expect(tt._import);
  this.expect(tt.parenL);
  if (!this.match(tt.string)) {
    throw this.unexpected(
      null,
      "Argument in a type import must be a string literal",
    );
  }
  node.argument = this.parseLiteral(this.state.value, "StringLiteral");
  this.expect(tt.parenR);

  if (this.eat(tt.dot)) {
    node.qualifier = this.tsParseEntityName(/* allowReservedWords */ true);
  }
  if (this.isRelational("<")) {
    node.typeParameters = this.tsParseTypeArguments();
  }
  return this.finishNode(node, "TSImportType");
}
```
So it looks like the `argument` has to be a `StringLiteral`, whereas the `qualifier` is a `TSEntityName`, aka [`Identifier | TSQualifiedName`](https://github.com/babel/babel/blob/738060ebfa7ac133d7dd6a2590835acaa08f15f3/packages/babel-parser/src/types.js#L1068). You're right that `argument` is mandatory but `qualifier` is optional.

bnjmnt4n:
Looking at https://github.com/babel/babel/pull/9309, `typeParameters` is an [optional `TSTypeParameterInstantiation` node instead of a `TSTypeParameterDeclaration` node](https://github.com/babel/babel/pull/9309/files#diff-0729be6aceed3c3d27b34e6226aa1509R414). Also, [`TSTypeQuery` should be updated to include `TSImportType`](https://github.com/babel/babel/pull/9309/files#diff-0729be6aceed3c3d27b34e6226aa1509R188)

benjamn:
@bnjmnt4n Great feedback! I will push a few more commits after I fix the unrelated build failures on `master`.

PR #330, opened 2019-04-22
Add TypeScript type alias declarations to scope

devongovett:
Flow type aliases were scope tracked correctly, but TypeScript type aliases were not. This should fix that.

I wasn't quite sure where to add a test for this. Let me know if/where I should add one.


benjamn:
@devongovett Thanks for working on this, and for using `ast-types` for `react-docgen`!

Can you either rebase against `master` now that I've fixed the tests, or open your branch so I can force-push to it?

If you have time to write a test, I would put it in `test/typescript.ts`, since this functionality is mostly TypeScript-specific, but you can model it after [similar `Scope` tests in `test/ecmascript.ts`](https://github.com/benjamn/ast-types/blob/2edd0e9f29e1aaf09e71ddc70b9897e54cbb204e/test/ecmascript.ts#L1074) (using `scope.lookupType` and `scope.declaresType` instead of `scope.lookup` and `scope.declares`).

devongovett:
Merged master and added tests.

benjamn:
Now available in `ast-types@0.12.4`. Thanks again!

PR #335, opened 2019-05-02
Add missing 'typeAnnotation' argument to TSAsExpression builder

unki2aut:
The original issue in `jscodeshift` https://github.com/facebook/jscodeshift/issues/314

### DIFF
--- a/def/typescript.ts
+++ b/def/typescript.ts
@@ -70,7 +70,7 @@ export default function (fork: Fork) {
 
   def("TSAsExpression")
     .bases("Expression", "Pattern")
-    .build("expression")
+    .build("expression", "typeAnnotation")
     .field("expression", def("Expression"))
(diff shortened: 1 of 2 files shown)


mdarveau:
I don't believe this is fixed.

```
j.tsAsExpression( j.memberExpression( j.identifier( "toto" ), j.identifier( 'titi' ) ), j.tsAnyKeyword() );
```
Gives me `no value or default function given for field "typeAnnotation" of TSAsExpression("expression": Expression)`

But this works fine:
```
j.tsAsExpression.from( {
    expression: j.memberExpression( j.identifier( "toto" ), j.identifier( 'titi' ) ),
    typeAnnotation: j.tsAnyKeyword(),
  },
);
```

I'm not sure where `tsAsExpression` is actually implemented, but it looks like it does not accept `typeAnnotation` as a second argument.

Thanks

PR #337, opened 2019-05-12
Stop making Expression inherit from Pattern.

benjamn:
Allowing any expression to serve as a binding pattern weakened the type system. It's time to deal with the (relatively small) consequences of ending that subtyping relationship.


def/core.ts
```diff
@@ -264,7 +263,7 @@ export default function (fork: Fork) {
         .bases("Expression")
         .build("operator", "left", "right")
         .field("operator", AssignmentOperator)
-        .field("left", def("Pattern"))
+        .field("left", or(def("Pattern"), def("MemberExpression")))
```
benjamn: The fact that a `MemberExpression` can appear on the left-hand side of an `AssignmentExpression` does not make it a binding pattern.

Issue #324, opened 2019-03-27
Simple example of how to use the builders and namedTypes in Typescript.


IAlwaysBeCoding:
Being new to Typescript I'm having a lot of problems trying to figure out how to use ast-types. I have been using ast-types with plain nodejs by following the example on the docs.

example
```nodejs

var n = require("ast-types").namedTypes;
var b = require("ast-types").builders;

var fooId = b.identifier("foo");

```

However, when I try to use it on Typescript I can't seem to figure out how to get the builders and namedTypes imported with Typescript.

I've tried these combinations

```nodejs
#This example only exports the types.
import {Builders, NamedType} from "ast-types"

#I tried this combination to get the builders, and no luck.
import * as astTypes from "ast-types"
console.log(astTypes.builders)

```

I really just want a simple easy 2-5 line example that shows me how to use ast-types in Typescript as I'm already doing it on Nodejs.

qiansc:
It seems that types and instance are not matched.
```
import  types from 'ast-types';
// type is right ,but types is undefined
// TypeError: Cannot read property 'builders' of undefined
console.log(types.builders);
```

PR #338, opened 2019-05-12
Export field values from namedTypes namespace to improve namedTypes ergonomics.

benjamn:
Although #300 vastly improved the behavior of `builders` by enabling autocompletion and type-checking of builder parameters, the `namedTypes` namespace remains somewhat difficult to use, as reported in #324 and elsewhere.

### ⚠️ The original plan of using declaration merging did not work because `namespace` declarations cannot be merged with value declarations in the way I naively imagined. Instead, this PR now models `namedTypes` as a TypeScript `namespace` that contains not only exported types but also exported fields, which are populated with appropriate values at runtime.

This PR is the beginning of an effort to take advantage of TypeScript's [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) abilities, which is relevant to `namedTypes` because the `namedTypes` identifier is not only (1) a `namespace` of `interface` types like
```ts
export namespace namedTypes {
  ...
  export interface IfStatement extends Omit<Statement, "type"> {
    type: "IfStatement";
    test: K.ExpressionKind;
    consequent: K.StatementKind;
    alternate: K.StatementKind | null;
  }
  ...
}
```
but also (2) an `interface` that contains `Type` fields for each AST node type:
```ts
export interface namedTypes {
  ...
  IfStatement: Type<namedTypes.IfStatement>;
  ...
}
```
and finally (3) an object that contains runtime values for all those properties, initially created [here](https://github.com/benjamn/ast-types/blob/a84563a26d85b238fccb464bc53a8065633e7c34/lib/types.ts#L813) and populated [here](https://github.com/benjamn/ast-types/blob/a84563a26d85b238fccb464bc53a8065633e7c34/lib/types.ts#L705-L709).

When you `import { namedTypes } from "ast-types"`, you expect to be able to use it in any of these three ways. Declaration merging is the magic that allows us to represent all three use cases using the same importable `namedTypes` identifier.

This will be a breaking API change for the `ast-types` package, since consumers will need to import `namedTypes` explicitly
```ts
import { namedTypes } from "ast-types"
namedTypes.IfStatement.assert(...);
```
rather than importing a default namespace and then accessing its `namedTypes` property
```ts
// This no longer works!
import types from "ast-types"
types.namedTypes.IfStatement.assert(...);
```


main.ts
```diff
@@ -8,22 +8,32 @@ import esprimaDef from "./def/esprima";
...
+  builders,
+  builtInTypes,
+  defineMethod,
+  eachField,
+  finalize,
+  getBuilderName,
+  getFieldNames,
+  getFieldValue,
+  getSupertypeNames,
+  namedTypes,
```
benjamn: This is the runtime object version of `namedTypes` (3).

main.ts
```diff
@@ -8,22 +8,32 @@ import esprimaDef from "./def/esprima";
 import babelDef from "./def/babel";
 import typescriptDef from "./def/typescript";
 import esProposalsDef from "./def/es-proposals";
-import { Omit } from "./types";
 import { ASTNode, Type, AnyType, Field } from "./lib/types";
 import { NodePath } from "./lib/node-path";
-import { NamedTypes } from "./gen/namedTypes";
-import { Builders } from "./gen/builders";
+import { namedTypes } from "./gen/namedTypes";
```
benjamn: This is both the `namespace` and `interface` versions of `namedTypes` (1, 2).

main.ts
```diff
@@ -38,18 +48,29 @@ const defs = [
...
+  builtInTypes,
+  defineMethod,
+  eachField,
   Field,
+  finalize,
+  getBuilderName,
+  getFieldNames,
+  getFieldValue,
+  getSupertypeNames,
+  namedTypes,
```
benjamn: Exporting `namedTypes` here combines all three versions together, though using a named export requires consumers to import `namedTypes` by name.

Issue #360, opened 2019-08-23
Variables declared in object destructuring aren't collected as bindings when using babel parser


RoystonS:
I was just trying out my very first bit of jscodeshift and got immediately stuck with an issue that my object destructuring assignments weren't appearing as bindings.  I've tracked it down to an issue here, but it's only an issue when using the babel/babylon parser.

e.g. with this code:
```
const { a, b } = something;
```
`ast-types` correctly reports both `a` and `b` as bindings when using the esprima parser, but not when using the babel parser.

I've added a unit test to cover this case, and I've also found a fix for it, which seems to be that the `scope` code is only looking for `Property` types when walking across `properties`, but it should also consider `ObjectProperty`.  I'm not 100% sure that this is the correct fix as I'm new to this codebase...

I'll raise an accompanying PR.

0xdevalias:
The seemingly related PR is merged, so I assume this issue should be resolved now?

- https://github.com/benjamn/ast-types/pull/361

PR #361, opened 2019-08-23
Scope issue with destructuring with babel parser

RoystonS:
This is a PR for issue #360.  It includes two commits:

1. Add a unit test to check that object destructuring is producing bindings correctly, across both esprima and babel parsers. (It currently fails for the babel parser.)
1. Add a fix for the babel parser, to include checking for `ObjectProperty` as well as just `Property`.  I don't know this code base at all so I may have fixed it in an entirely awful way. (I've noticed other places sometimes checking for the existence of `namedTypes.XXX` before using them. Is that necessary here?)

### DIFF
--- a/test/ecmascript.ts
+++ b/test/ecmascript.ts
@@ -1209,6 +1209,31 @@ describe("scope methods", function () {
     assert.deepEqual(names, ["x", "xyDefault", "xyNamespace", "z"]);
   });
 
+  describe("getBindings should work with destructuring operations", function() {
+    const code = `
+function aFunction(arg1, { arg2 }) {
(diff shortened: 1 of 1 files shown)


gavrix:
The same happens with tsx parser (when used with jscodeshift) 
is there anything preventing this to be merged?

gavrix:
@benjamn would you be able to take a look?

benjamn (approved):
@RoystonS The same bug fix (supporting `ObjectProperty`) was added in a later PR #540 (sorry for missing this!), but I'm very happy to add your regression test to the suite. Sorry for the long delay.

PR #365, opened 2019-09-23
Add ImportExpression node type (ES2020)

mkubilayk:
ES2020 grammar has `ImportExpression` node type for dynamic `import()`. Based on [ESTree spec](https://github.com/estree/estree/blob/master/es2020.md#importexpression), this PR adds support for it in `ast-types`.


mkubilayk:
Hi @benjamn - I'd appreciate if we can get this in. Do you have time to take a look?

GeeWizWow:
Hi, any plans to merge this CL ?

benjamn:
Please note that Babel already had an `Import` node that was used as the `callee` of a `CallExpression` to represent dynamic `import(...)` calls. That was changed without much discussion in https://github.com/estree/estree/issues/197, which was shipped along with https://github.com/estree/estree/pull/198. I must be missing some context around how changes like these get approved and shipped, but for now `ast-types` can just support both versions.

mkubilayk:
Thanks for merging this @benjamn!

PR #371, opened 2020-02-12
Add Flow type imports to scope types, not bindings

motiz88:
Previously, Flow `import type` specifiers would erroneously be counted as creating value bindings in a scope. With this change they are correctly added to `scopeTypes` instead.


blaine-stripe:
Just ran into this as well.

@benjamn is there something more we can do to get this upstreamed?

PR #376, opened 2020-04-02
Update Flow definitions

### LINKED ISSUE #370: Flow TypeParameter is missing type parameter defaults
^ causes recast/eslint/jscodeshift to miss things like this:
```
type X<T =  NEVER_TRAVERSED> = {}
```

will follow up with a fix

gkz:
This pull request updates the Flow definitions. While Flow does not have a formal AST specification for Flow specific additions/changes, we do maintain a set of AST types that have not yet upstreamed (https://github.com/facebook/flow/blob/master/packages/flow-parser/test/custom_ast_types.js) - this PR upstreams the Flow specific ones.

Changes:
- Now that the `es2020` defs were added, fork on that rather than the previous latest, `es7`
- Changed `var`s to `const`s
- Added `SymbolTypeAnnotation`, `BigIntTypeAnnotation`, and `BigIntLiteralTypeAnnotation` defs
- Added `ObjectTypeIndexer` def
- Made the `name` field of `FunctionTypeParam` optional
- Added a `default` field to `TypeParameter`
- Made `supertype` field of `OpaqueType` optional
- Fixed up the definition of `DeclareOpaqueType`
- Added additional things that can be exported in `DeclareExportDeclaration`
- Added `predicate` fields to function related defs
- Added `"typeof"` as option to `importKind` field of `ImportDeclaration` 
- Added a variety of Flow enums related defs
- And updates the version of Flow parser used to the latest

Note, there are some additional changes we need to upstream around export related nodes, however these need to be moved/added to the JS definitions first before we can update them here. Will do that in a future PR.

Closes #370 
Closes #345 

Tested with `npm test`


Daniel15:
This would be really useful for my jscodeshift use case. Is it possible to merge this?

gkz:
@benjamn I'd really appreciate it if you were able to look at this :)

AuthorityFX:
Why is there no movement on this?

gkz:
Thanks for merging! Yes, the latest parser (0.132) is great - just wasn't updated from when I initially made the PR

PR #377, opened 2020-04-02
Update JavaScript definitions

gkz:
This is an extensive refactor and update of the JS definitions. It is based on the ESTree spec (https://github.com/estree/estree) while keeping existing legacy behaviour.

Changes:
- Adds a variety of missing definitions
- Create files for es2016-es2019 (delete `es7` file, which had a mix of definitions from a mix of years), and move every definition/change to the correct year. For example, if a change was added in a later standard (e.g. catch clause param can be `null` in ES2019), the definition in `core.ts` should not reflect this, but the new file `es2019.ts` updates the definition
- Move definitions in `babel-core`that were actually JS definitions, to the correct file
- Created a `core-operators.ts` file which defines arrays of the various operators. This makes it easy (and safer) to later introduce a new operator in later versions of the standard 
- There are some changes in terms of the `ImportSpecifier` nodes from the es6 definition - now what we `.build` follows the ESTree standard and what was defined in the `babel-core` definitions - `imported` and `local` rather than `id` and `name`


gkz:
Thanks for merging!

PR #378, opened 2020-04-21
fix!: unify babel/esprima specifier types, making `local` a required field

sodatea:
This reverts the change landed at https://github.com/benjamn/ast-types/commit/a4cd3e9f43667856f29b2067485d97a96a994fc8

As https://github.com/jquery/esprima/pull/1149 was merged long ago,
now the esprima AST is compiliant with the ESTree spec, `id` and `name`
fields now longer exist. It's time to unify the specifier definitions.

Documentation also available at https://esprima.readthedocs.io/en/4.0/syntax-tree-format.html#import-declaration


sodatea:
@benjamn I've rebased and re-generated the types

Issue #366, opened 2019-10-02
TS2590: Expression produces a union type that is too complex to represent.


dbaum:
I think tsc 3.6.3 is having trouble with the union of values in builtInTypes:

types.ts:411:5 - error TS2590: Expression produces a union type that is too complex to represent.

411     builtInTypes[name] = type;
        ~~~~~~~~~~~~~~~~~~

SnarkBoojum:
I'm using tsc 3.7.4  and I see the same problem...

SnarkBoojum:
Same issue with tsc 3.8.2.

SnarkBoojum:
Is there something I can do to help pinpoint the issue and be done with it?

SnarkBoojum:
Same issue with tsc 3.8.3.

SnarkBoojum:
Sifting through my list of pending issues with my packages, I see I can't update ast-types in Debian because it doesn't work with recent tsc.

SnarkBoojum:
Still no luck with ast-types 0.13.3 : doesn't work with tsc 3.8.3.

SnarkBoojum:
Just retried with tsc 3.9.2 and ast-types 0.13.3 : same error.

SnarkBoojum:
Same issue with tsc 3.9.7.

PR #384, opened 2020-08-14
Upgrade to latest typescript, fix TS2590 and TS2339 errors in typescript 3.5+

fdintino:
Updates typescript to 3.9.7. Also updates `@babel/parser` to the minimal minor version that supports assertion signatures 
(7.7.7).

Fixes benjamn/ast-types#366

Apologies for the wall of text to follow; hopefully it will be useful to some future person trying to resolve a similar issue.

I have to confess that I don't fully understand why `defBuiltInType` causes the error described in #366 
 (TS2590: Expression produces a union type that is too complex to represent), but I spent way too long trying to do so. In typescript 3.5, the compiler started enforcing some types more aggressively (This comment on a somewhat related issue opened for the TypeScript repo was helpful: https://github.com/microsoft/TypeScript/issues/33130#issuecomment-526373717) . Some combination of the pseudo-self-relational typing for the ten properties of `BuiltInTypes` (e.g., initializing `builtInTypes` as an empty object and using `typeof isString` before `isString` gets defined) and the fact that `Type` in `lib/types.ts` is a union of four types meant that the typescript checker's internal representation of the type would need a union of 4^10 types. It calculates the size of this union and throws the above error if it exceeds 100000. I gave up trying to figure whether it's actually necessary for the typescript checker to expand all possible types in this way (in which case this would actually be a typescript bug). In any case, the simplest solution to fix the compiler error here was to assign the `builtInTypes` object all at once, and allow typescript to construct the type of that object based on the inferred return types of `defBuiltInType<T>`.

In addition to TS2590 in #366, the typescript 3.5 compiler was throwing another error inside the conditional in `defBuiltInType`:

```
lib/types.ts:413:35 - error TS2339: Property 'constructor' does not exist on type 'T'.

413     if (example && typeof example.constructor === "function") {
                                      ~~~~~~~~~~~

lib/types.ts:414:35 - error TS2339: Property 'constructor' does not exist on type 'T'.

414       builtInCtorFns.push(example.constructor);
                                      ~~~~~~~~~~~
```

This was remedied by restricting the type of `example` to the finite set of possible types. I slightly reworked the function signature so that `name` parameter could be used for the generic type variable.

My motivation for this PR was to be able to use ast-types with a later version of typescript. I initially did the work of supporting an upgrade of both typescript and babel to the latest version, but the babel upgrade involved many more changes. In the interest of causing fewer conflicts with other open PRs, I tried to limit things as much as possible. That still required an update to `@babel/parser`, which I did to the earliest minor release that has compatibility with typescript assertion signatures.


wbt:
> Apologies for the wall of text to follow; hopefully it will be useful to some future person trying to resolve a similar issue.

It was, thank you!

Issue #383, opened 2020-08-03
Support `ChainExpression` type for optional chaining


mjhenkes:
# Feature request or bug?

babelParser@7.11.0 changed how optional chaining gets represented in an AST. Previously the optional chain was represented as an `OptionalMemberExpression`, now the optional chain is broken into two nodes a `ChainExpression` and a `MemberExpression`.

I asked babel if this was a breaking change and they clarified that this is an intentional change. See [Babel issue #11908](https://github.com/babel/babel/issues/11908)

I stumbled upon this while trying to use react-doc-gen against a component with optional chanining and the ast-types library that it uses fails when it tries to understand the 'ChainExpression' type. See [react-doc-gen issue #463](https://github.com/reactjs/react-docgen/issues/463)

I'm currently working around this issue by locking in the projects babel dependencies in the package.json

For example:

```diff
  "devDependencies": {
-    "@babel/core": "^7.5.0",
+    "@babel/core": "7.10.5",
+    "@babel/parser": "7.10.5",
```

However this isn't a good long term fix. I would be ideal if ast-types was updated to support the `chainExpression` type.

ethancrook99:
I made a PR for a possible fix [here](https://github.com/benjamn/ast-types/pull/399).

mjhenkes:
Thank you @ethancrook99 and @benjamn ❤️

PR #399, opened 2020-08-25
Added ChainExpression type

ethancrook99:
Added a new `ChainExpression` type to line up with how babelParser@7.11.0 handles optional chaining. I left `OptionalChainExpression` but marked it as deprecated for projects that use older babel versions.

Meant to fix [this issue](https://github.com/benjamn/ast-types/issues/383)


ethancrook99:
@benjamn Can I get a review on this?

ethancrook99:
@benjamn Sure, I'll get started on those changes!

ethancrook99:
@benjamn When adding the `expression` field, do I add the `ChainElement` type label as a string inside a `def()`? Like this--
```  
def("ChainExpression")
    .bases("Expression")
    .build("object", "property", "computed", "optional")
    .field("optional", Boolean, defaults["true"])
    .field("expression", def("ChainElement"));
```

It seems like that's the pattern used for other types but I can't tell for sure. I pushed the changes, will you let me know if I missed something?

ethancrook99:
@benjamn I changed the supertype of `CallExpression` and `MemberExpression` to `ChainElement`, but it's causing a lot of tests to fail because those two types are no longer a subtype of `Expression` and aren't related to `ExpressionKind`. Do you know of a way to resolve that conflict?

info-bit:
Are you planning to merge this fix anytime soon? We're stuck with a yarn resolution for babel in the package to support react-docgen.

benjamn (approved):
@ethancrook99 Thanks for continuing to work on this, and sorry for leaving you to figure out the key insight about multiple inheritance. That's one of the reasons I built this whole `def` system in the first place, instead of using ordinary JS prototypal inheritance, since prototypes allow for only single inheritance (with good reason, of course, but you really want multiple inheritance, or something like traits, for an AST type system like this).

ethancrook99:
Thanks for finishing that up and getting it merged!

Issue #326, opened 2019-03-29
Spread operator in function arguments (RestElement) not treated as declaration


strobox:
I found that using ast-types@0.12.3
`function(e,...t) {`
Only *e* appears in bindings.

MuTsunTsai:
I just experienced the same issue; any news? I'm using version 0.14.1.

MuTsunTsai:
See PR #409

PR #409, opened 2020-09-08
Fix issue #326

MuTsunTsai:
See #326


benjamn (approved):
Makes sense—thanks @MuTsunTsai!

PR #416, opened 2020-09-14
Set `tslib` as dev dependency

### LINKED ISSUE #415: Question: Should `tslib` be a dependency or a dev dependency?
I noticed `tslib` was added as a dependency. I'm not super familiar with typescript but it seemed to stick out as the only non dev dependency and it seems like it's only used in internal typescript.
https://github.com/benjamn/ast-types/blob/master/package.json#L40

Added here:
https://github.com/benjamn/ast-types/commit/03d6467df0d9edb1abf13da741e3f81e847feee6

mjhenkes:
Assuming `tslib` should be a dev dependency, heres a PR! If not, no harm done, feel free to close. 😄 

Closes #415

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1124,7 +1124,8 @@
     "tslib": {
       "version": "2.0.1",
       "resolved": "https://registry.npmjs.org/tslib/-/tslib-2.0.1.tgz",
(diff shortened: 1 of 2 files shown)


benjamn:
@mjhenkes Let us know if you run into any problems with `tslib`, since it's not strictly necessary, but it does greatly reduce the size of the TS-generated code.

PR #455, opened 2020-12-07
Added support for block scoping

### LINKED ISSUE #154: path.scope not accurate for const and let
Variables declared with `let` and `const` are block-scoped, not function-scoped. Yet `path.scope` is the same for both blocked-scoped variables and function-scoped variables declared with `var`. The following

```
var recast = require("recast");                                                   
var types = require("ast-types");                                                 
                                                                                  
var ast = recast.parse(`                                                          
  function f() {                                                               
    if (true) {                                                                
      const x = 3;                                                             
    }                                                                          
  }                                                                            
`);                                                                            
                                                                                                                                                                                                           
types.visit(ast, {                                                                
  visitVariableDeclaration: function(path) {                                      
    console.log(path.scope.node.type);                                            
    return false;                                                                 
  }                                                                               
});
```

Outputs "FunctionDeclaration", but I think it should be "BlockStatement".

davidnormo:
An attempt to resolve #154

### DIFF
--- a/README.md
+++ b/README.md
@@ -437,6 +437,18 @@ contains `path.node`. See [scope.ts](lib/scope.ts) for its public
 interface, which currently includes `.isGlobal`, `.getGlobalScope()`,
 `.depth`, `.declares(name)`, `.lookup(name)`, and `.getBindings()`.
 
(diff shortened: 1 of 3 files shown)


davidnormo:
@benjamn the failing tests relate to the catch block bindings no longer being hoisted. My changes have not covered hoisting because the ECMA specs talk about hoisting as a runtime behaviour. Is that right? Or do I need to cover `var` hoisting in all blocks?

davidnormo:
@benjamn sorry to ping you again. Let me know what you think when you get time

PR #467, opened 2020-12-23
fix(scope): Support ClassDeclarations without identifier

danez:
Class declarations might not have an identifier when used in export default.

Fixes reactjs/react-docgen#472


benjamn (approved):
Sorry for the long wait @danez!

PR #540, opened 2021-05-07
scopes: support ObjectProperty in addPattern()

henryqdineen:
This should fix an issue that we ran into using jscodeshift while using the `renameTo()` functionality. I noticed that the issue does not happen when using the default `babel` parser ([AST Explorer snippet](https://astexplorer.net/#/gist/8a4e8879cff65dbeda21098bd7776d17/a483790fbbe20e050f5ff9ad63687227628cd257)) but the issue does occur with the `tsx` or `babylon` parsers ([AST Explorer snippet](https://astexplorer.net/#/gist/8a4e8879cff65dbeda21098bd7776d17/4cc31b43fea756876ffab2e9d9e6d6b96404f83f)). 

The issue is that the Babel 7 configurations used in the `tsx` and `babylon` parsers use the `ObjectProperty` node type instead of `Property`, which is used by `flow` and older versions of Babel.

### DIFF
--- a/lib/scope.ts
+++ b/lib/scope.ts
@@ -318,8 +318,10 @@ export default function scopePlugin(fork: Fork) {
       namedTypes.AssignmentPattern.check(pattern)) {
       addPattern(patternPath.get('left'), bindings);
 
-    } else if (namedTypes.ObjectPattern &&
-      namedTypes.ObjectPattern.check(pattern)) {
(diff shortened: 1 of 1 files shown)


benjamn (approved):
Thanks @henryqdineen—and sorry for the long delay!

PR #664, opened 2021-11-29
Refactor `{Binary,Assignment,Logical}Operators` by specification year

benjamn:
This PR adds an `operators` subdirectory to the `def` directory, with separate files for each specification year that modified the operators used for binary, assignment, and logical expressions.

Instead of evaluating these definition files just once, they now participate in the `Fork` system so that they can be reimported/reevaluated multiple times.


main.ts
```diff
@@ -38,24 +31,14 @@ const {
...
-
   // Feel free to add to or remove from this list of extension modules to
   // configure the precise type hierarchy that you need.
-  es6Def,
-  es2016Def,
-  es2017Def,
-  es2018Def,
-  es2019Def,
-  es2020Def,
+  esProposalsDef,
```
benjamn, line 36: Since `esProposalsDef` imports all previous specification years and the core definitions, we can/should use only `esProposalsDef` here, so types will not be prematurely finalized by loading previous years first. Note that removing the previous defs does cause some harmless reordering of generated types.

PR #840, opened 2022-09-26
Make `assert` method an assertion function in TS

Andarist:
This is available since TS 3.7:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions

It could be considered a breaking change. Note that DefinitelyTyped supports officially only the last 2 years of TS releases and that makes 4.1 the last supported version there: https://github.com/DefinitelyTyped/DefinitelyTyped#support-window

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -27,7 +27,7 @@ abstract class BaseType<T> {
 
   abstract check(value: any, deep?: Deep): value is T;
 
-  assert(value: any, deep?: Deep): value is T {
+  assert(value: any, deep?: Deep): asserts value is T {
     if (!this.check(value, deep)) {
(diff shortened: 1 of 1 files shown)


davbrito:
I think this would be a great addition to the project. Any particular reason why it's not been merged yet?

As for the breaking change, what about adding an extra method called `assertType`? That way, people who are currently expecting the true return can still use `assert` without any problem.

acidoxee:
I'd like to second this, the `.check`/`.assert` distinction is already native to the lib, it's just missing the proper typings. It's a bummer having either to nest the code in `if` statements (to benefit from the current, limited type narrowing) or having to resort to unsound casts 🙏

PR #886, opened 2022-12-11
Restore whole-program parsing/validation tests, with necessary `def` updates and repository maintenance

benjamn:
This should address @eventualbuddha's comment https://github.com/benjamn/ast-types/pull/327#issuecomment-1335967133, though of course reenabling those whole-program parsing and validations tests meant fixing the ones that were failing.

~There are still two tests failing when checking `ast.errors` against `expected.errors`. If I can't figure out what's going on with those test fixture files tomorrow, I'll just skip parsing/validating them.~  Skipped! e0ae66e99c38ecb777a08eb293c6c6b1eeec8dc3

New definitions added:
- [`TSInstantiationExpression`](https://github.com/benjamn/ast-types/pull/886/commits/118e3778d6674f7113613a44a547aa1b983983d5)
- [`TSTypeCastExpression`](https://github.com/benjamn/ast-types/pull/886/commits/4a014e0ddd5a288fe117468080f0a6514af3d80d)
- [`TSSatisfiesExpression`](https://github.com/benjamn/ast-types/pull/886/commits/4a014e0ddd5a288fe117468080f0a6514af3d80d)
- [`ClassAccessorProperty`](https://github.com/benjamn/ast-types/pull/886/commits/7e43b0fe1de211df392f7c6defaeebfe0976f9d8)


package.json
```diff
@@ -56,7 +56,7 @@
     "flow-parser": "0.195.0",
     "glob": "8.0.3",
     "mocha": "^9.1.3",
-    "recast": "0.21.5",
+    "recast": "0.23.0-pr-1247",
```
eventualbuddha, line 59: Intentional?

eventualbuddha (approved):

src/test/typescript.ts
```diff
@@ -13,59 +13,106 @@ import { NodePath } from "../node-path";
...
     throw error;
   }
 
+  if (files.length < 10) {
+    throw new Error(`Unexpectedly few **/input.ts files matched (${
+      files.length
+    }) in ${
+      babelTSFixturesDir
+    }`);
+  }
```
benjamn, line 42: Hopefully this check will prevent the scenario where the `glob` pattern suddenly fails to match any files (such as the upstream project deciding to bulk-rename all the `input.ts` files).

package.json
benjamn: Intentional but temporary: 3db28be204d49c317d63888b3aa4e72e5638e760

Issue #509, opened 2021-03-19
Mozilla parser API link is dead


aureooms:
The link https://developer.mozilla.org/en-US/docs/SpiderMonkey/Parser_API is dead. It is used in the first paragraph of the README.

AndyOGo:
Different version can be found in an archive, e.g.
http://web.archive.org/web/20140802223108/https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Parser_API

And this presentation by Michael Ficarra
https://speakerdeck.com/michaelficarra/spidermonkey-parser-api-a-standard-for-structured-js-representations

PR #936, opened 2023-04-04
fix: update SpiderMonkey Parser API dead link

Benjscho:
Update the dead link to the SpiderMonkey parser API to a web archive link of the same page from 2014.


Benjscho:
This fixes [issue 509](https://github.com/benjamn/ast-types/issues/509)

Issue #823, opened 2022-08-29
[Proposal] we should provide a function to make a fork from @babel/types


jedwards1211:
I've been thinking, sometimes Babel nodes and fields are missing from the defs in `ast-types`, it's naturally going to lag behind what they're adding support for.  Why not build the defs from the information in `@babel/types`?  Here's what I have so far.  Seems to work well:

```ts
import typesPlugin from 'ast-types/lib/types'
import * as t from '@babel/types'
import { mapValues } from 'lodash'
import fork from 'ast-types/fork'
import { Fork } from 'ast-types/types'
import { parse } from '@babel/parser'
import nodePathPlugin from 'ast-types/lib/node-path'

function babel(fork: Fork) {
  const types = fork.use(typesPlugin)
  const { builtInTypes, Type } = types
  const { def, or } = Type

  fork.use(nodePathPlugin)

  def('Node').field('type', builtInTypes.string)

  function tryConvertValidate(validate: any, node?: any): any {
    if (validate.type) {
      switch (validate.type) {
        case 'any':
          return {}
        case 'string':
          return builtInTypes.string
        case 'boolean':
          return builtInTypes.boolean
        case 'number':
          return builtInTypes.number
        case 'null':
          return builtInTypes.null
        case 'undefined':
          return builtInTypes.undefined
      }
    }
    if (validate.chainOf) {
      for (const elem of validate.chainOf) {
        const converted = tryConvertValidate(elem)
        if (converted) return converted
      }
    }
    if (validate.each) {
      return [convertValidate(validate.each)]
    }
    if (validate.oneOfNodeTypes) {
      return or(...validate.oneOfNodeTypes.map((type: string) => def(type)))
    }
    if (validate.oneOf) {
      return node?.optional
        ? or(...validate.oneOf, null)
        : or(...validate.oneOf)
    }
    if (validate.shapeOf) {
      return mapValues(validate.shapeOf, (value) =>
        convertValidate(value.validate, value)
      )
    }
    if (validate.oneOfNodeOrValueTypes) {
      return or(
        ...validate.oneOfNodeOrValueTypes.map((type: string) =>
          /^[A-Z]/.test(type) ? def(type) : convertValidate({ type })
        )
      )
    }
  }

  function convertValidate(validate: any, node?: any): any {
    const converted = tryConvertValidate(validate, node)
    if (!converted) {
      throw new Error(
        `couldn't determine field def for validate: ${JSON.stringify(validate)}`
      )
    }
[...]

PR #937, opened 2023-05-07
feat: add overrideDefs function

jedwards1211:
Solving https://github.com/benjamn/recast/issues/1283 would be difficult because
it imports ast-types everywhere, and I would have to replace those references
with passing an ast-types fork as new function parameters all over the place.

This is the pragmatic solution, it would allow me to just replace all of the
ast-types exports with my own fork built from `@babel/types` (see
https://github.com/benjamn/ast-types/issues/823).  So anywhere that imports
them would get my definitions that are up-to-date with babel.

Building defs from `@babel/types` or other parsers' metadata is the only
sustainable option for the community to have a reliable codemod system built on recast.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -7,11 +7,11 @@ import babelDef from "./def/babel";
 import typescriptDef from "./def/typescript";
 import { ASTNode, Type, AnyType, Field } from "./types";
 import { NodePath } from "./node-path";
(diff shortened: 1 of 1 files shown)


jedwards1211:
Ugh this won't work the best either...I would have to make sure I override the defs before I import recast

PR #945, opened 2023-09-27
Update docs to point to src/def subdirectory

ryanwilsonperkin:


benjamn (approved):
Thanks @ryanwilsonperkin!

PR #946, opened 2023-10-14
Fix scope lookup with optional catch binding

### LINKED ISSUE #944: Missing catch binding breaks scope API
Using `v0.16.1`.

After parsing the following:

```js
try {} catch { id; }
```

When walking the AST and then looking up the scope at `id` an exception is thrown:

```
 null does not match type Pattern
    at BaseType.assert (\node_modules\ast-types\lib\types.js:15:19)
    at addPattern (\node_modules\ast-types\lib\scope.js:256:28)
    at scanScope (\node_modules\ast-types\lib\scope.js:136:17)
    at Sp.scan (\node_modules\ast-types\lib\scope.js:109:13)
    at Sp.getBindings (\node_modules\ast-types\lib\scope.js:114:14)
    at Context.visitIdentifier (\ast.js:30:32)
    at Context.invokeVisitorMethod (\node_modules\ast-types\lib\path-visitor.js:283:51)
    at PVp.visitWithoutReset (\node_modules\ast-types\lib\path-visitor.js:158:32)
    at visitChildren (\node_modules\ast-types\lib\path-visitor.js:204:25)
    at PVp.visitWithoutReset (\node_modules\ast-types\lib\path-visitor.js:167:20)
```

Steps to reproduce:

```js
const astTypes = require("ast-types");

const {builders: b, visit} = astTypes;

`try {} catch (e) { void x }`;

const main = b.file(b.program([
    // try {}
    b.tryStatement(b.blockStatement([]),
    // catch
       b.catchClause(
        /* param: */ null, // b.identifier("e"), // <- adding a catch binding avoids the bug
        /* guard: */ null,
        // {
        b.blockStatement([
            // void x
            b.expressionStatement(
                b.unaryExpression("void", b.identifier("x"))
            )
       ]))
    // }
    )
]));

visit(main, {
    visitIdentifier(path) {
        console.log(`at ${path.node.name}`); // 'at x'
        console.log(path.scope.getBindings()); // <- throws
        return false;
    }
});

```

andersk:
- Fixes #944.

### DIFF
--- a/src/scope.ts
+++ b/src/scope.ts
@@ -197,7 +197,7 @@ export default function scopePlugin(fork: Fork) {
       }
     }
     if (ScopeType.check(node)) {
-      if (namedTypes.CatchClause.check(node)) {
+      if (namedTypes.CatchClause.check(node) && node.param !== null) {
         // A catch clause establishes a new scope but the only variable
         // bound in that scope is the catch parameter. Any other
         // declarations create bindings in the outer scope.


kj800x:
:+1: We ran into this bug when running a codemod on our codebase and spent a few hours debugging it today. We're going to try to patch-package this fix into our ast-types dependency. It would be great to get this merged upstream if it works.

